PR #59, opened 2013-01-03
Bringing back old-style com_fixedTic, adding Carmack's reverse as a compile option

meklu:
These commits bring back the pre-patch behaviour where setting com_fixedTic to -1 unlimits the framerate while still somewhat keeping everything in sync. (com_fixedTic 1 would make the physics engine go super fast in addition to the framerate, and this is undesirable.) This should help people with, say, 120Hz monitors.

Another feature that was added was the option to compile the engine to use Z-fail in its shadow rendering. This is off by default.

### DIFF
--- a/neo/CMakeLists.txt
+++ b/neo/CMakeLists.txt
@@ -30,6 +30,7 @@ option(D3XP			"Build the d3xp game code" ON)
 option(DEDICATED	"Build the dedicated server" OFF)
 option(ONATIVE		"Optimize for the host CPU" OFF)
 option(SDL2			"Use SDL2 instead of SDL1.2" OFF)
+option(ZFAIL			"Use Carmack's reverse" OFF)
 
 if(NOT CMAKE_SYSTEM_PROCESSOR)
 	message(FATAL_ERROR "No target CPU architecture set")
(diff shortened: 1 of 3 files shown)


shoober420:
Any chance for the com_fixedTic command to allow -1 again for 60+ FPS?

revility:
Picked up a 144hz monitor.  The higher fps thing is like everyone keeps saying.  You don’t notice it until you go back to something lower.  Agreed that it would be cool to have unlocked frames option even if it’s considered experimental or something.

shoober420:
It’s not about seeing a difference, but feeling the difference.

DanielGibson:
btw, `com_fixedTic -1` has been back since 1.5.0

I'll keep this open in case people are interested in the Carmack's Reverse thing that's not gonna get merged.

DanielGibson:
TIL that this fucking patent is finally gonna expire this year (https://twitter.com/EricLengyel/status/1151652012619993088)

https://patents.google.com/patent/US6384822B1/en says October 13th
I guess unless it turns out that the patent for some reason runs longer than that after all, the changes for "Carmack's Reverse" can be merged then :)

Rinnegatamante:
Nump, patent expired two days ago for this

DanielGibson:
I just committed Z-Fail support - based on https://github.com/ljbade/doom3.gpl/commit/d4de024341e79e0ac1dfb54fb528859f8ccea605 by @ljbade which looked more correct than this PR.

I also enhanced it to use glStencilOpSeparate() (if available), which reduces the numbers of draw calls for shadows - that code is based on @raynorpat's https://github.com/raynorpat/Doom3/blob/2933cb554587aea546c2df1fd/neo/renderer/draw_stencilshadow.cpp

Both are enabled by default, but can be disabled with the `r_useCarmacksReverse` and `r_useStencilOpSeparate` CVars in case you wanna make comparisons in performance or how it looks.
On my desktop machine (Ryzen 2700X + Geforce GTX 970) and on my Raspberry Pi 4B it didn't really make a noticeable difference though.

PR #97, opened 2014-06-13
Make the game launchable on linux with a system installation

chungy:
As it is, the engine won't find the game data if the following is true:
1. The user is not in the directory containing the Doom 3 base folder
2. The binary is not installed in the same directory either (eg, in /usr/bin)
3. The user tries to simply launch the game by running the binary


DanielGibson:
yeah, this was indeed a bug and I finally fixed it in c333a463733d40f26044e09f590b85d10bdf6226

if in the executable dir there's no base/ dir, it falls back to LINUX_DEFAULT_PATH (by default `/usr/local/games/doom3/`)

PR #113, opened 2015-04-23
Adding OpenBSD compatibility

fmwviormv:
This project will compile on my system (OpenBSD 5.6-stable GENERIC.MP#333 amd64) with these few patches.

### DIFF
--- a/neo/CMakeLists.txt
+++ b/neo/CMakeLists.txt
@@ -99,6 +99,9 @@ include_directories(${VORBISFILE_INCLUDE_DIR})
 find_package(OpenAL REQUIRED)
 include_directories(${OPENAL_INCLUDE_DIR})
 
+find_package(OpenGL REQUIRED)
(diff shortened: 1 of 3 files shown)


DanielGibson:
Sorry for the late reply, I totally missed this pull request :-/
(Didn't github at some point list pull requests in the issues?)

Anyway, is the change to CMakeLists.txt really necessary?
dhewm3 doesn't link against libGL but loads all needed GL functions through SDL with `SDL_GL_GetProcAddress()` (see https://github.com/dhewm/dhewm3/blob/master/neo/renderer/RenderSystem_init.cpp#L613-L619 https://github.com/dhewm/dhewm3/blob/master/neo/sys/glimp.cpp#L326-L330 and https://github.com/dhewm/dhewm3/blob/master/neo/renderer/qgl_proc.h)

So I *think* that using SDL's include dirs and linking against it should suffice?

DanielGibson:
fixed by #153

PR #116, opened 2015-06-20
Add support for brazilian keyboard layout

franmagneto:
Now brazilian players can open the console


DanielGibson:
sorry for ignoring the pull request for so long, must have missed it :-/

the patch for brazilian keyboard support is merged now, thanks!

DanielGibson:
now that I think about it again.. is it really smart to make the ' and " key console key? then you can't write quotes in the game (neither in console nor in chat...)

andre-d:
Why is this required anyway?

DanielGibson:
to detect the "console key", which is the one under escape and has different keycodes in different layouts

andre-d:
Right but why can that key not just be the key code rather than the
physical character?  And can we not use an sdl text input event to get the
actual Unicode character typed if we really wanted to?

DanielGibson:
the keycode is mapped to the doom3 keycode before, which corresponds to ascii for ascii keys.

we do use the actual unicode character for console input, but that doesn't fix anything: you type `"` (or whatever the key is in your layout), you'll get an `"` doom3 event which opens/closes the console and a `"` unicode event which is written.
but as the console is opened/closed at the same time the writing doesn't really happen.

franmagneto:
I always use this key to open the console on FPS games, and this is the same that you use (the one under escape), and I don't remember having issues with it in any game. Maybe I never tried to write a " or ' character in console, haha

andre-d:
Right, but SDL has a scancode which tells you which physical key was pressed, if we want it to be consistent why are we not just using the scancode and it will not matter what layout your computer thinks it is in?

andre-d:
If you want "the key under escape" the scancode should just be SDL_SCANCODE_GRAVE should it not?

andre-d:
There is likely just some key bit of information I am missing about keyboard layouts or what we are trying to accomplish though

DanielGibson:
as I wrote before, on Brazilian keybaords that key below escape is used for `"`.  
so if someone with that kind of keyboard wants to write a `"` in the console, it closes instead.

DanielGibson:
brazilian support is still in but the "console key" can now be disabled with `in_ignoreConsoleKey 1` (you can still open the console with shift+esc then)

PR #130, opened 2015-10-10
make idlib build with Visual Studio 14

leffmann:
the changes you suggested will allow VS14 to build idlib correctly.


DanielGibson:
I squashed the commits into one and commited it.

Thanks!

PR #131, opened 2015-10-11
more fixes for VS 14

leffmann:
About the changes in codec.cpp: MSVC defines `HUGE_VAL` as `INFINITY`, while idMath does `#undef INFINITY`, presumably because it wants you to use its own `idMath::INFINITY` instead.

Feel free to pull and squash.

### DIFF
--- a/neo/cm/CollisionModel_load.cpp
+++ b/neo/cm/CollisionModel_load.cpp
@@ -1887,7 +1887,7 @@ static int CM_FindSplitter( const cm_node_t *node, const idBounds &bounds, int *
 						continue;
 					}
 					// find the most centered splitter
-					t = abs((bounds[1][type] - dist) - (dist - bounds[0][type]));
+					t = idMath::Fabs((bounds[1][type] - dist) - (dist - bounds[0][type]));
 					if ( t < bestt ) {
 						bestt = t;
 						*planeType = type;
@@ -1907,7 +1907,7 @@ static int CM_FindSplitter( const cm_node_t *node, const idBounds &bounds, int *
 						continue;
 					}
 					// find the most centered splitter
-					t = abs((bounds[1][type] - dist) - (dist - bounds[0][type]));
+					t = idMath::Fabs((bounds[1][type] - dist) - (dist - bounds[0][type]));
 					if ( t < bestt ) {
 						bestt = t;
 						*planeType = type;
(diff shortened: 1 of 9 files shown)


DanielGibson:
I'll wait until everything builds before merging.

DanielGibson:
replacing `abs()` with `idMath::Fabs()` may not yield the same behavior, afaik `abs()` implicitly casts to int first. not sure if that behavior actually was intended and if something breaks if it's changed

leffmann:
You're right. The `std::abs`, if that's what is effectively being called, should be overloaded, but I'm not sure either what the real intention is.

Replacing HUGE_VAL with idMath::INFINITY should be fine given what the code does, and since it's defined not as infinity but as 1e30.

With the above commits everything builds, and seems to play fine, so feel free to pull if you think it looks OK.

DanielGibson:
`std::abs()` has an overload for float, see http://www.cplusplus.com/reference/cmath/abs/

but I don't think that `std::abs()` was called - there's no `using namespace std;` in the code, so  `abs()` - should have called C's `abs()` (http://www.cplusplus.com/reference/cstdlib/abs/) which can't be overloaded, because C doesn't support function overloading.

DanielGibson:
ok, arguing with standard behavior when talking about MSVC is kinda silly.
I just tested the following Code with MSVC6, MSVC2010 Express on WinXP and clang++ and g++ on Linux:
```
#include <cmath>
#include <cstdlib>
#include <cstdio>

int main()
{
	float f = -2.45f;
	float f2 = abs(f);
	printf("abs(%f) = %f\n", f, f2);
	return 0;
}
```

When built with MSV6 or (clan)g++ it prints `abs(-2.450000) = 2.000000`.
When built with MSVC2010 it prints `abs(-2.450000) = 2.450000`.

So somewhere between MSVC6 and 2010 these strange overloads for float were introduced, i.e. not just `float std::abs(float)` but also `float abs(float)` in the global namespace - WTF?!

Not sure what doom3 originally was built with - certainly not MSVC6, maybe 7.0 or 7.1 (2003)?

But either way, binaries of the code built with MSVC2010 or newer should have been sufficiently tested to assume that replacing `abs()` with `idMath::Fabs()` should be safe.

DanielGibson:
merged (after removing the special casing for MSVC2015 with the _fileno() stuff)

DanielGibson:
Thanks for the patches!

PR #134, opened 2015-10-22
added autofilling r_vidmodes via SDL and automatic use max resolution

cjayho:
Don't know why DHEWM developers so many years didn't do such simple work. We have SDL, which can list available video modes. I changed static array of video modes to dynamic idList, and made automatic fill this idList with available video modes. By default game uses maximum available resolution.
There are two implementations of videomodes list generation - for SDL1 and SDL2. At this time SDL2 version supports only default display.

### DIFF
--- a/neo/renderer/RenderSystem_init.cpp
+++ b/neo/renderer/RenderSystem_init.cpp
@@ -40,6 +40,8 @@ If you have questions concerning this license or the applicable additional terms
 
 #include "renderer/tr_local.h"
 
+#include <SDL.h>
+
 // Vista OpenGL wrapper check
 #ifdef _WIN32
 #include "sys/win32/win_local.h"
@@ -54,7 +56,7 @@ const char *r_rendererArgs[] = { "best", "arb2", NULL };
 idCVar r_inhibitFragmentProgram( "r_inhibitFragmentProgram", "0", CVAR_RENDERER | CVAR_BOOL, "ignore the fragment program extension" );
 idCVar r_useLightPortalFlow( "r_useLightPortalFlow", "1", CVAR_RENDERER | CVAR_BOOL, "use a more precise area reference determination" );
 idCVar r_multiSamples( "r_multiSamples", "0", CVAR_RENDERER | CVAR_ARCHIVE | CVAR_INTEGER, "number of antialiasing samples" );
-idCVar r_mode( "r_mode", "5", CVAR_ARCHIVE | CVAR_RENDERER | CVAR_INTEGER, "video mode number" );
+idCVar r_mode( "r_mode", "0", CVAR_ARCHIVE | CVAR_RENDERER | CVAR_INTEGER, "video mode number" );
 idCVar r_displayRefresh( "r_displayRefresh", "0", CVAR_RENDERER | CVAR_INTEGER | CVAR_NOCHEAT, "optional display refresh rate option for vid mode", 0.0f, 200.0f );
 idCVar r_fullscreen( "r_fullscreen", "0", CVAR_RENDERER | CVAR_ARCHIVE | CVAR_BOOL, "0 = windowed, 1 = full screen" );
 idCVar r_customWidth( "r_customWidth", "720", CVAR_RENDERER | CVAR_ARCHIVE | CVAR_INTEGER, "custom screen width. set r_mode to -1 to activate" );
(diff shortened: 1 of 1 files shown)


DanielGibson:
> Don't know why DHEWM developers so many years didn't do such simple work.

For several reasons.. the first is, that this is only possible because of `R_GetVidModeValsString()` and `R_GetVidModeListString()` which I added only a few weeks ago, see https://github.com/dhewm/dhewm3/commit/c0e8e20628b423dd6d8640565cab95c4e30dfce9

Apart from that your solution isn't viable for several reasons:
* Those resolutions from SDL are only supported fullscreen resolutions. The hardcoded list is also useful for windowed mode
* The mode index (which ends up as r_mode) is not consistent. It depends on what and how many resolutions are returned by SDL and might change when changing desktop resolution, updating the GPU driver, updating SDL, adding a display and probably the moon phase.

I fear that with how the Doom3 system menu works, we can't do much better than the hardcoded list of resolutions we currently have (and it was already a big improvement that I could add the widescreen resolutions to that list without touching the menu definitions).

Yamagi:
I highly suspect that SDL_GetDisplayMode() doesn't work correctly with non trivial multiscreen setups. At least on unixoid platforms with X11 and maybe Wayland the rectangular base geometry of the frame buffer breaks things if the real display geometry has more than 4 corners. Nvidia added the ability to fake / overwrite / hack the mode list and to disable displays at resolution change for exactly that reasons.

neo/renderer/RenderSystem_init.cpp
```diff
@@ -550,6 +534,19 @@ idStr R_GetVidModeValsString()
...
 /*
 ==================
+R_AddVideoMode
+
+This function adds videomode to videomodes list
+*/
+
+void R_AddVideoMode( int width, int height )
+{
+	r_vidModes.Append( *( new vidmode_t( width,	height ) ) );
```
DanielGibson: btw, this leaks memory.

cjayho:
i just wondering - is there ability to add any features to main dhewm3? i think maintainer is too lazy even to merge features made by another people :(.

kortemik:
I am sorry to tell you but the reasons given are valid. Although the idea is nice, the implementation was less working, and it doesn't fit all the custom gui-menus out there.

But if you want to keep working on it, there are plenty of other corners to make smooth.

DanielGibson:
WTF.

> i just wondering - is there ability to add any features to main dhewm3? i think maintainer is too lazy even to merge features made by another people :(.

and your editing to 
>  think it will be better to use own branch instead of getting declines with stupid reasons.

isn't much better.

Your initial post ("Don't know why DHEWM developers so many years didn't do such simple work.") was already rude and nevertheless I replied and pointed out why the problem isn't as easily solvable as you assumed.

But instead you go on insulting me as "lazy" and denote the reasons I gave you (which you obviously didn't even try to comprehend) as "stupid".

Please also note that I do this in my freetime without any payment. So whatever I'm doing here, "lazy" is no appropriate label.

Feel free to do you own branch with buggy "features" and memory leaks, I don't care, but I won't merge that kind of code here (if I did, dozens of other people would turn up and complain about the bugs :stuck_out_tongue:).

cjayho:
I just want to know why you're talking me about memory leaks, and valgring did not?! that fullfilling of idList is made only once (on running subroutine R_InitOpenGL). Where did you notice that memory leak?

kortemik:
I see your patch calling "new" and no "delete" therefore it leaks.

Please refer to http://www.amazon.com/Effective-Specific-Improve-Programs-Designs/dp/0321334876

DanielGibson:
`r_vidModes.Append( *( new vidmode_t( width,	height ) ) );`
is equivalent to
```c++
vidmode_t* tmp = new vidmode_t( width, height );
r_vidModes.Append( *tmp );
```
r_vidModes is an `idList<vidmode_t> r_vidModes;`, i.e. it contains vidmode_t objects, *not* pointers to them that may be deleted later (which is why you dereferenced the object)
so `r_vidModes.Append( *tmp );` will create a copy of `tmp` in `r_vidModes` and  `tmp` will leak.

cjayho:
DanielGibson, ok, thanks for the answer. I'll try to fix this.

cjayho:
so this variant:

```
typedef class vidmode_s {
	public:
		int	width, height;

		vidmode_s(){};
		vidmode_s( int w, int h )
		{
			width = w;
			height = h;
		}
} vidmode_t;

idList<vidmode_t*> r_vidModes;

/* --- */

void R_AddVideoMode( int width, int height )
{
	r_vidModes.Append( new vidmode_t( width,	height ) );
	common->Printf( "New video mode: \"Mode %d: %dx%d\"\n", r_vidModes.Num()-1, width, height );
}

```

is better? or the same?

DanielGibson:
not really better, you'd have to call delete on everything in r_vidModes; furthermore the code accessing r_vidModes would have to be adjusted.

the easiest/best fix for this particular problem would be to just don't call `new()`, but do `r_vidModes.Append( vidmode_t( width, height ) );` instead.

But please note that the memory leak is only one small problem, the other three issues I stated above (SDL resolutions only for fullscreen, inconsistencies with what r_mode values mean, might break mods) remain and are not really fixable considering how the doom3 system menu works.

motorsep:
@DanielGibson Heh, after reading this stuff, it makes me wonder why people at D3 forums make me look like a bad guy :)

@cjayho Why don't you fork dhewm3, add whatever you need to add there, make it work flawlessly, test it and get users' feed back and then present it to dhewm3 devs ?

cjayho:
DanielGibson, about delete of vidModes i did also think, so added such code to the subroutine idRenderSystemLocal::Shutdown before ShutdownOpenGL() :

```
	for ( int i = 0; i < r_vidModes.Num(); i++ ) {
		common->Printf( "Deleted mode %d: %dx%d\n", i, r_vidModes[i]->width, r_vidModes[i]->height );
		delete r_vidModes[i];
	}

	r_vidModes.Clear();
```

btw, sorry for my words, i was wrong :(

DanielGibson:
deleting it in `idRenderSystemLocal::Shutdown` /might/ work, yeah. but using an idList with objects instead of pointers is the better choice anyway, just don't call new when adding elements.

Another thing to keep in mind is that `R_InitOpenGL()` is **not** only called at startup, but after each vid_restart - but `idRenderSystemLocal::Shutdown()` isn't. so after a vid_restart everything will be in there twice, most probably. Lots of things must be regarded for changes like this ;-)

> btw, sorry for my words, i was wrong :(
I'm glad you see that now :-)

cjayho:
> SDL resolutions only for fullscreen

vanilla doom3 had square resolutions from 640x480 to 1280x1024 in both modes and worked in both modes without any problems. My code makes these modes too, so i don't see any problems with this.

> inconsistencies with what r_mode values mean, might break mods

i can add r_SDLMode cvar, if you afraid to break compatibility.

>  and are not really fixable considering how the doom3 system menu works

i think we need modification of game menu scripts. i know how menu switch works.

cjayho:
> Another thing to keep in mind is that R_InitOpenGL() is not only called at startup, but after each vid_restart

hm. checked, and it really is. Made subroutine

```
void R_ClearModes( void )
{
	for ( int i = 0; i < r_vidModes.Num(); i++ ) {
		common->Printf( "Deleted mode %d: %dx%d\n", i, r_vidModes[i]->width, r_vidModes[i]->height );
		delete r_vidModes[i];
	}

	r_vidModes.Clear();
}
```

and added to the end of idRenderSystemLocal::Shutdown() and to the beginning of the R_InitOpenGL().

DanielGibson:
> vanilla doom3 had square resolutions from 640x480 to 1280x1024 in both modes and worked in both modes without any problems. My code makes these modes too, so i don't see any problems with this.

Does it? Where? What resolutions SDL gives you at least depends on your GPU driver, SDL version, what options (randr etc) SDL is compiled with (I observed these behaviors when implementing getting the resolution list for rbdoom3bfg). So these resolutions would have to be added manually to be sure.

> i can add r_SDLMode cvar, if you afraid to break compatibility.

The problem with r_SDLMode would be that r_mode is hardcoded in the .gui

In theory it would be nice if one could display the list of SDL resolutions if fullscreen is enabled in the menu, and the static list if it's not. But I doubt this can be done, because of the inflexibility of the menu.
It'd also be nice if one could select which display to use so we could get the list of resolutions for it in SDL2 - but this also isn't possible without changing the proprietary mainmenu.gui (most probably).

So the best we could end up with is always displaying the current list of hardcoded resolutions + the list of resolutions from SDL.
I'm not sure if this makes a difference in 99% of the cases, i.e. if anyone has resolutions there that aren't in the list anyway - and those few people could as well use r_customWidth/Height (if there are a few common resolutions missing they can always be added to the hardcoded list)

PR #135, opened 2015-11-02
More mouse buttons + more keyboard keys

twolife:
Hi, here are 2 usability patches :
 - up to 8 mouses buttons with SDL2 >= 2.0.4 (you already know the sad story with << 2.0.4 )
 - support more keyboard keys (at least now i can map all the weapons on my belgian keyboard)

I don't know if it's the right way to do all this, but at least "it works for me"

### DIFF
--- a/neo/sys/events.cpp
+++ b/neo/sys/events.cpp
@@ -628,6 +628,18 @@ sysEvent_t Sys_GetEvent() {
 					mouse_polls.Append(mouse_poll_t(M_DELTAZ, -1));
 				break;
 #endif
+			default:
+#if SDL_VERSION_ATLEAST(2, 0, 0)
+				// handle X1 button and above
+				if( ev.button.button <= 8 ) // doesn't support more than 8 mouse buttons
+				{
+					int buttonIndex = ev.button.button - SDL_BUTTON_LEFT;
+					res.evValue = K_MOUSE1 + buttonIndex;
+					mouse_polls.Append( mouse_poll_t( M_ACTION1 + buttonIndex, ev.button.state == SDL_PRESSED ? 1 : 0 ) );
+				}
+				else
+#endif
+				continue; // handle next event
 			}
 
 			res.evValue2 = ev.button.state == SDL_PRESSED ? 1 : 0;


DanielGibson:
The mouse button patch looks good.

I'm not sure about the keyboard changes - why do you have to remove all the AUX keys?
Why don't you have mappings for the added keys in `keynames[]`? What is even shown in the menu if you try to bind the new keys there? (I think the menu code uses `keynames[]` to get the string to display)

twolife:
>  I'm not sure about the keyboard changes - why do you have to remove all the AUX keys?

in KeyInput.h there is a comment that says " // this better be < 256!" , so i removed all the K_AUX that seems completely unused to keep it < 256  while adding what was necessary

> Why don't you have mappings for the added keys in keynames[]?

because it's not necessary. (but please don't ask me why ;)

> What is even shown in the menu if you try to bind the new keys there?

it shows 'é' , '§' , 'è' , 'ç'

DanielGibson:
> it shows 'é' , '§' , 'è' , 'ç' 
and this is expected? those aren't even ASCII chars (except for '$' of course)..
I think I'd be more comfortable if they had ASCII names for binding etc.

twolife:
> and this is expected? those aren't even ASCII chars (except for '$' of course)

that what i expects yes, all the other keys around ( '&', '(', '!', ')' ) are printed just like that. if i remember correctly when i played the original .exe from IDSoftware back in the days on windows (with the same keyboard layout) that was also like that.

(PS: it's not a '$', it's a '§' , something i never used for another purpose than selecting a gun in an FPS ;)

DanielGibson:
(yeah, I know it's '§', it was just a typo, on German keyboard they're next to each other. lawyers use it to reference articles of laws. or something like that)

Anyway, maybe it's fine to do it like this.

However: `case 167:` etc - not good, I wouldn't trust it to work for SDL1.2 vs SDL2.
Isn't there an SDLK_* constant for those keys?

DanielGibson:
Regarding `SDLK_*`: at least for SDL2 it seems like there are no small constants over 126 `'z'` (for all the other constants its `$sdl2_scancode | (1 << 30)` which is really big.

So there shouldn't even be key events with keycodes like 167, 224, 231, ...
And if there are I doubt they're portable between different operating systems.

DanielGibson:
Hmm could you tell me, what SDL2 keycodes and scancodes those keys generate for you? And what operating system you're using and the exact version of SDL2?

This might help: https://gist.github.com/DanielGibson/12041b57a540cbbfbaf9

twolife:
The good news is that it's working with SDL1.2.

> what SDL2 keycodes and scancodes those keys generate for you?

SDL_KEYDOWN: Keycode: é (233) Scancode: 2 (31)
SDL_TEXTINPUT: é
SDL_KEYUP: Keycode: é (233) Scancode: 2 (31)
SDL_KEYDOWN: Keycode: è (232) Scancode: 7 (36)
SDL_TEXTINPUT: è
SDL_KEYUP: Keycode: è (232) Scancode: 7 (36)
SDL_KEYDOWN: Keycode: ç (231) Scancode: 9 (38)
SDL_TEXTINPUT: ç
SDL_KEYUP: Keycode: ç (231) Scancode: 9 (38)
SDL_KEYDOWN: Keycode: § (167) Scancode: 6 (35)
SDL_TEXTINPUT: §
SDL_KEYUP: Keycode: § (167) Scancode: 6 (35)


> And what operating system you're using and the exact version of SDL2?

Ubuntu 15.10 on amd64.
SDL is a HG snapshot from a week or two, 2.0.4~9893

DanielGibson:
thanks for the information :)
So the Scancodes are the same as on any keyboard - they're the numbers that QWERTY keyboards have in that row. This is expected, but good to know it's actually true for AZERTY, too.

Could you test on Windows and/or OS X as well?

twolife:
> Could you test on Windows and/or OS X as well?

Windows 7 amd64 :
SDL_KEYDOWN: Keycode: 2 (50) Scancode: 2 (31)
SDL_TEXTINPUT: é
SDL_KEYUP: Keycode: 2 (50) Scancode: 2 (31)
SDL_KEYDOWN: Keycode: 7 (55) Scancode: 7 (36)
SDL_TEXTINPUT: è
SDL_KEYUP: Keycode: 7 (55) Scancode: 7 (36)
SDL_KEYDOWN: Keycode: 9 (57) Scancode: 9 (38)
SDL_TEXTINPUT: ç
SDL_KEYUP: Keycode: 9 (57) Scancode: 9 (38)
SDL_KEYDOWN: Keycode: 6 (54) Scancode: 6 (35)
SDL_TEXTINPUT: §
SDL_KEYUP: Keycode: 6 (54) Scancode: 6 (35)
Received SDL_QUIT - bye!

I don't have any Apple hardware to test on OS X

DanielGibson:
Thanks a lot!
As you see, the keycodes aren't even consistent between Windows and Linux :-/

I'm currently writing an E-Mail to the SDL mailing list to ask for advice. I think they should fix it and I wonder that apparently no one has complained about this before (at least not loud enought ;-))

DanielGibson:
So far I didn't get interesting feedback on the SDL mailinglist, apart from the information that OSX seems to behave like Linux.

I created a bugreport at SDL: https://bugzilla.libsdl.org/show_bug.cgi?id=3188

However, I thought about a workaround: We could just check the SDL scancodes and always map `SDL_SCANCODE_1, _2, ..., _9, SDL_SCANCODE_0` to `1, 2, ..., 9, 0` - like on QWERTY.
That way we don't have to rely on undefined keycodes on SDL and we don't have to introduce new `K_*` constants in Doom3.

Do you see any problem with that?

The code in neo/sys/events.cpp, `Sys_GetEvent()` would look somehow like this:
```c++
			// fall through
		case SDL_KEYUP:
#if !SDL_VERSION_ATLEAST(2, 0, 0)
			key = mapkey(ev.key.keysym.sym);
			if (!key) {
				// .. unchanged ...
			}
#else // SDL2 and newer - note that SDL1.2 doesn't really have scancodes
			// workaround for AZERTY-keyboards, which don't have 1, 2, ..., 9, 0 in first row:
			// map to those keycodes anyway
			if(ev.key.keysym.scancode == SDL_SCANCODE_0)
			{
				key = '0';
			}
			else if(ev.key.keysym.scancode >= SDL_SCANCODE_1 && ev.key.keysym.scancode <= SDL_SCANCODE_9)
			{
				// note that the SDL_SCANCODEs are SDL_SCANCODE_1, _2, ..., _9, SDL_SCANCODE_0
				// while in ASCII it's '0', '1', ..., '9' => handle 0 and 1-9 separately
				// (doom3 uses the ASCII values for those keys)
				key = '1' + ev.key.keysym.scancode - SDL_SCANCODE_1;
			}
			else
			{
				key = mapkey(ev.key.keysym.sym);
			}
			if(!key) {
			//  ... the rest is unchanged ...
```

DanielGibson:
@twolife could you test if the latest master code works for you regarding your first key row? :-)

(When using SDL2) it should pretend it's 1, 2, ..., 9, 0 now for keybindings etc, while text input (console, chat, ..) should still work and behave like you'd expect from your keyboard layout.

I tested it a bit (with just changing my keyboard layout in the XFCE keyboard settings) and it seemed to work, but I'd like confirmation from someone who actually owns such a keyboard.

twolife:
It seems to work, thanks a lot for the fix :-)
seems odd that nobody has ever hitting that limitation in libSDL before & that you have to workaround that in the game; let's hope that the resulting discussion will be productive ^^

DanielGibson:
Yep, I also surprised that there was nothing in the SDL bugzilla about this (apart from some very old bug for SDL1 that mentioned AZERTY but failed to articulate the problem).

However it was funny that not even two weeks after I mentioned the problem on the SDL mailing list someone else posted about the same problem with OSX.

And some months ago we had the same bug reported for Yamagi Quake II, but at that time I didn't realize it's an SDL bug and thought we'd just have to map some additional `SDLK_*` keycodes to new Quake2 `K_*` constants (they're basically the same as in doom3): https://github.com/yquake2/yquake2/issues/81

Anyway, I hope that it'll eventually get some attention from SDL developers. If they'd at least define the expected behaviour, so I could start writing a patch at least for Linux/X11 (and potentially Win32, if the current behavior - which is like what we do here in dhewm3 - isn't desired)

DanielGibson:
So, about the support 8 mouse buttons thing: I think it should be for SDL2 2.0.0 and newer.
I know the buttons are messed up on Linux/X11 before 2.0.4, but afaik they  worked fine on Windows and probably OSX and other operating systems.

Furthermore I think that the default case should have a `continue;` for SDL1 or unknown mouse buttons on SDL2, so we don't generate garbage events there for unknown mouse buttons (as we already did before your patch. oops.)

DanielGibson:
oh sorry, I somehow missed that you've updated the pull request (github didn't send an email)

I'll try it out and merge it soon, thanks!

DanielGibson:
finally merged it, thanks a lot :-)

PR #160, opened 2017-03-14
Fix stack overflow in SSE code

turol:
This was checking the wrong variable so when count was < 4 it was writing
past the end of buffer and potentially breaking the stack.


DanielGibson:
Thank you!
How did you find it?

turol:
AddressSanitizer

DanielGibson:
cool

PR #161, opened 2017-03-14
Fix idStr self-assignment

turol:
Was calling memcpy with overlapping parameters which is inefficient,
undefined behavior and Valgrind complained about it.

### DIFF
--- a/neo/idlib/Str.h
+++ b/neo/idlib/Str.h
@@ -519,6 +519,10 @@ ID_INLINE char &idStr::operator[]( int index ) {
 #pragma GCC diagnostic pop
 
 ID_INLINE void idStr::operator=( const idStr &text ) {
+	if (&text == this) {
+		return;
+	}
+
 	int l;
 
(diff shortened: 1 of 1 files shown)


DanielGibson:
Does it actually happen? If so, there's probably a bug in the caller, wouldn't make much sense..

turol:
https://github.com/dhewm/dhewm3/blob/master/neo/framework/Common.cpp#L2558
The code calls StripFileName on a variable and assigns the return value back to it. StripFileName returns reference to "this", resulting in self-assignment. The comment seems to indicate this is intentional.

DanielGibson:
ok, that was a workaround for a compilerbug.. I wonder if this breaks the workaround, after all it's an inline function and the compiler might notice that it'll do nothing and again optimize everything out..
but it's not like gcc 4.7 is very relevant anymore, so I'll just take that risk ;)

DanielGibson:
thanks a lot for the patch :)

PR #162, opened 2017-03-14
Fix read from uninitialized part of buffer

turol:
According to "man 2 readlink": "readlink() does not append a null byte to buf."
I have no idea how this worked up to now.


DanielGibson:
holy shit you're right, how did this work? maybe readlink() adds `\0` after all?

turol:
No it doesn't, Valgrind sees an uninitialized read in strlen() inside idStr::operator=()
https://github.com/dhewm/dhewm3/blob/master/neo/idlib/Str.cpp#L177
Also, that operator= checks for self-assignment.

DanielGibson:
anyway, thanks again, keep 'em coming! ;)

turol:
Watch out what you wish for, you just might get it :)

DanielGibson:
hey, I meant pull requests, not bug reports ;-P

PR #180, opened 2017-11-28
Some spelling error fixes found during Debian build

coldtobi:
- s/allready/already
- s/Uknown/Unknown
- s/thier/their

Those are found during the build for the new Debian package (lintian informational messages)
Would be great if you could merge them to silence those.

Many thanks!
tobi (your dhewm3 package maintainer in Debian)


DanielGibson:
thanks!

PR #182, opened 2018-01-20
Improve spelling and grammar in README

Calinou:
This pull request also renames OS X to macOS, which is the official name since 10.12 Sierra.


DanielGibson:
thanks!

Issue #192, opened 2018-06-21
UIs sometimes unresponsible


Yamagi:
Sometimes UIs only register clicks after a random key on the keyboard is pressed. A good place to reproduce this in playing video disks through the PDA. Every one and then video doesn't start, even if the "play" button is clicked several times.

Yamagi:
This is somewhat strange:

* The event is registered and - when testing with the PDA - the first frame of the requested video is played. The video stucks at the second frame.
* The bug is not reproduceable with all videos (and likely GUIs in general but I'm not sure on that).
* Not every key has the same probability to get things going. Closing the PDA with <tab> always starts the video playback.

DanielGibson:
merged your fix so this can be closed :)

PR #194, opened 2018-07-03
Lower release build optimizations from -O3 to -O2.

Yamagi:
In the last few weeks I've played again through the game and kept an eye
on small oddities. And there're a lot of them. For example some GUIs and
videos getting stuck after the first frame (issue #192) or being unable
to protect the guy with the lamp in Alpha Labs 3. Some digging proved
that most - if not all - of these problems are caused by the compilers
optimization level. When build with -O2 both g++ 8.1 and clang 6.0.0 are
producing working code. g++ 8.1 with -O3 has some small, hard to notice
oddities, clang 6.0.0 with -O3 shows a lot of them. Since there's not
measurable difference between -O3 and -O2 just go down to the later:

```
x doom_o3.txt
+ doom_o2.txt
+----------------------------------------------------------------------+
|                                                               +      |
|                                      +                        *      |
|x                                     *           x            *      |
|                 |_____________________|___A______M__A_________M___|_||
+----------------------------------------------------------------------+

    N        Min           Max        Median           Avg        Stddev
x   5        173           178           177         176.4     2.0736441
+   5        176           178           178         177.2     1.0954451
```


DanielGibson:
awesome, thank you!

Nable80:
It's important to remember that most probably the oddities are caused by UB (undefined behavior) in code and not by the compiler bugs (which do exist but they are extremely rare in production versions). I wonder what can be discovered with ``-fsanitize=address``, ``-fsanitize=memory``, ``-fsanitize=undefined`` (both GCC and Clang support these flags now), etc.

DanielGibson:
as long as using the less aggressive optimizations of -O2 suffice to make it work I won't spend much time on this
(I have used ASan on dhewm3 in the past though and only found some new[]/delete missmatches)

DanielGibson:
I mean, using all those sanitizers in a recent clang/gcc version might be interesting and fixing any issues found makes sense if that's no too painful, so feel free to go for it - PRs are welcome :)
(Note: we already build with -fno-strict-aliasing and this is not gonna change)
But I'm not too confident that all the (sometimes subtle) issues caused by -O3 can be found/fixed this way, so I prefer to keep using -O2 to be (a bit more) safe - and apparently it makes no difference for performance anyway.

Nable80:
I understand your point, it's completely valid. I just wanted to say that it's hard to be sure that compiler has caused the problem, most probably it uncovered something problematic.
It's nice to see that ``-O2`` made it work after all.

DanielGibson:
I'm quite sure it's not a real compiler bug (after all it affects both gcc and clang), but writing 100% conforming code is almost impossible at a large scale, and compilers tend to invent new interpretations of UB for new optimizations so it's a moving target anyway - and changing old code like this to be conformaning is probably even harder..
I think it's quite common to avoid -O3 because it tends to break things; I don't remember why dhewm3 used it

Yamagi:
Some parts of the source (for example the math lib at idlib/math) have been highly optimized by hand, maybe even up to the point that the code exploits UB to run faster. And at least the SSE implementations of some math functions in Simd_SSE.cpp have made problems before, causing strange behavior like clipping against brush edges on a planar surface.

Even if we're able to find such problems in the code, fixing them always imposes a high risk of changing there behavior in edge cases or with extreme values, imposing a high risk of breaking existing game data.

PR #196, opened 2018-07-26
Fix spelling errors (II)

coldtobi:
seems so that I've applied the previous patch only partially to my branch... This PR has the remaining spelling errors in my Debian patch.. Sorry about the extra work.

Tobi

### DIFF
--- a/neo/tools/compilers/roqvq/roq.m
+++ b/neo/tools/compilers/roqvq/roq.m
@@ -49,7 +49,7 @@ blah
 	return lastFrame;
 }
 
-- (BOOL)scaleable
+- (BOOL)scalable
 {
 	return [paramFileId isScaleable];
 }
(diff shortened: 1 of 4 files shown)


neo/tools/compilers/roqvq/roqParam.cpp
```diff
@@ -108,8 +108,8 @@ void roqParam::InitFromFile( const char *fileName )
 			readarg++;
 			continue;
 		}
-// scaleable
-		if (token.Icmp( "scaleable") == 0) {
+// scalable
+		if (token.Icmp( "scalable") == 0) {
```
DanielGibson: this looks like a very bad idea. apparently some kind of input file is expected to have this typo, changing this here will break parsing.

neo/tools/compilers/roqvq/roq.m
```diff
@@ -49,7 +49,7 @@ - (BOOL)isLastFrame
 	return lastFrame;
 }
 
-- (BOOL)scaleable
+- (BOOL)scalable
```
DanielGibson: no idea what this objective C file is really used for, but I'm not a fan of fixing typos in actual code.

(comments and error message strings should be ok, everything else seems risky)

coldtobi:
Revised patch with new commit.
NOTE: I accidentially force-pushed this, please make sure to throw away any local branch you might have.

Issue #193, opened 2018-06-21
Quick saving interrupts certains sound


Yamagi:
If a quick save is written certain sounds interrupt. This easy to reproduce with playing an audio log throug the PDA an hitting the quick save button. Or in low oxygen areas some background sound stop.

svdijk:
This is probably the same issue as #137.

Yamagi:
A possible fix is here: https://github.com/Yamagi/dhewm3/tree/alfix It needs some testing if there side effects.

Issue #137, opened 2015-11-10
Voice playback stops when quicksave is initiated


AMDmi3:
When I quicksave, currently played voice fragment (such as NPC speaking, or radio transmission) is silent until it ends. E.g. npc lips are still moving/transmission UI indication is still active, but no sound is heard.

AMDmi3:
Update: this actually affects all sounds, not just voice

svdijk:
I've checked this againt the original Doom3 1.3.1 binary, and this doesn't occur there, sounds keep playing there when I quicksave.

Note that the issue in dhewm3 is a bit peculiar: a sound stops playing when quicksaving _only_ when the sound is being played for the first time while quicksaving. E.g. when I start a specific PDA recording, then quicksave, then recording stops playing. However if I then start the same recording again, I can quicksave all I want, the recording keeps playing correctly. The same goes for other recordings and NPCs.

Maybe some caching issue?

PR #199, opened 2018-08-19
Fix entering the menu or saving the game stopping some sounds.

Yamagi:
If an OpenAL source runs out of samples it transisions into state
AL_STOPPED. That happens if we're entering the menu (which switches
to another soundworld) and when saving the game (because the game
blocks for some milliseconds). Work around this by adding a new
field 'stopped' to the channel state and use that to determine if
a sound was stopped. And not AL_STOPPED like before.


DanielGibson:
Thank you!

This should fix #193 and #137

PR #207, opened 2018-10-20
Spelling fixes -- Revised patch after comments

coldtobi:


DanielGibson:
Thank you!

By the way, I think there will be a new release in the foreseeable future - not exactly sure when, but I'll hopefully release RC1 in the next two weeks, to hopefully get people to test it a bit before the proper release.

PR #208, opened 2018-11-05
More complete idWindow::CleanUp

zinal:
Fixed idWindow::CleanUp to avoid references to already-freed items in idWindow::ops and idWindow::updateVars. Showed itself when trying to run with missing d3xp/pak000.pk4.


DanielGibson:
the commit is in master now, thanks a lot and sorry it took so long!

PR #218, opened 2019-01-03
Demo version support

gabrielcuvillier:
This is the very minimum changes needed for Dhewm work with D3 demo data

Basically, if we can't load the mainmenu gui, we check for the demo_mainmenu one + a few fixes here and then.

There is a lot of warnings while running the game, but demo is playable.

### DIFF
--- a/neo/d3xp/Weapon.cpp
+++ b/neo/d3xp/Weapon.cpp
@@ -29,6 +29,7 @@ If you have questions concerning this license or the applicable additional terms
 #include "sys/platform.h"
 #include "framework/DeclEntityDef.h"
 #include "framework/DeclSkin.h"
+#include "framework/Session.h"
 
 #include "gamesys/SysCvar.h"
 #include "ai/AI.h"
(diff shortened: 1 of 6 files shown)


DanielGibson:
Thanks a lot!

I won't merge it like this, because modifying Session.h (and particularly the `idSession` class) breaks the game API and ABI, which would break all the existing mods ported to dhewm3 for the recent 1.5.0 release (=> I'd have to modify the SDK as well and all the ported mods, and existing binaries wouldn't work anymore.. too messy).

However, this is still very valuable because it shows me what logical changes are needed - I just need to find a way to do them without breaking the API :-)

BTW, you did the changes in both game/ and d3xp/ - doesn't the demo only use/need game/ ?

gabrielcuvillier:
Ok, no issue. Well maybe the d3xp changes were not necessary :)

DanielGibson:
I've built upon your work in my d3demo branch: https://github.com/DanielGibson/dhewm3/tree/d3demo
It's still a bit WIP but seems to work very well - I could play the whole demo without any problems.

The only thing you missed was `idWeapon::EnterCinematic()` which also needs a check before calling something on `WEAPON_NETFIRING`.

gabrielcuvillier:
I merged you changes, this is working. I wonder why D3 demo support was not done before, as this is really simple ;)

PR #221, opened 2019-01-13
We need to link libexecinfo for the backtrace() functions on FreeBSD.

Yamagi:
Like the commit message says, -lexecinfo is required for FreeBSD now that the backtrace() functions are used.


DanielGibson:
thanks a lot!

PR #224, opened 2019-01-15
Compilation fix on 64 bit macOS. Upgrade min SDK version

Klezstyle:
Upgrade CMakeLists.txt minimum SDK version for compiling in macOS to v.10.9 "Mavericks" for 64 bit builds.
Until now, make failed with this error:

warning: include path for stdlibc++ headers not found; pass '-std=libc++' on the command line to use the libc++ standard library instead [-Wstdlibcxx-not-found]
In file included from /Juegos/dhewm3/neo/idlib/bv/Bounds.cpp:29:
/Juegos/dhewm3/neo/sys/platform.h:185:10: fatal error: 'cstddef' file not found
#include <cstddef>
         ^~~~~~~~~
1 warning and 1 error generated.


DanielGibson:
this `<cstddef>` thing is super weird - this has been a standard C++ header since C++98 (according to http://www.cplusplus.com/reference/clibrary/), which should be old enough even for Apple..

Issue #148, opened 2016-07-18
No text on BigEndian


tlosm:
Hi devs on bigendian text are not displayed 
if you need some logs or more please ask 
attached an image
![screenshot from 2016-07-18 20-18-51](https://cloud.githubusercontent.com/assets/18098090/16925513/e89967ee-4d24-11e6-9d49-39bb03590347.png)

DanielGibson:
uhh.. no idea, sorry. I can't reproduce the problem because I don't have any big endian hardware, and thus I can't debug it :-(

In case anyone else can debug this, what operating system and hardware are you using?

tlosm:
Hi Daniel, Im using Ubuntu Mate 16.10 on AmigaOne X5000 Freescale P5020 (dual 2ghz)with RadeonHD 6570 2GB , but i have the Same issue on a Quad G5 with Ubuntu Mate 15.10 with Radeon HD 4650 512mb . A friend of mine same issue on P6T 6870 1GB (all this machine are big endian) 
If you have a good horse powered machine you can use qemu-system-ppc with ubuntu mate emulated not fast but working .

DanielGibson:
Does qemu support OpenGL now? (Haven't used it in years.. not since Open Source VirtualBox came out for Linux, I think)

tlosm:
Yes thru virgl3d and with passthru. but for now not in the emulated hw only in virtualized. they made big steps in last year.

DanielGibson:
so that doesn't help me much..

I guess you're using a 64bit OS and 64bit dhewm3 binaries?
If so, can you build and run 32bit binaries (like x86 binaries on amd64)? Do they have the same bug?
If this is not possible, do you or one of your friends have a PPC G4 system or something like that with a 32bit OS where you can test?

(Generally, Doom3 should support PPC/big endian, at least it originally worked on OSX/PPC, so I wonder if it's a 64bit PPC problem. Of course it is entirely possible that we introduced the bug in the changes we did to the original source, but testing this might narrow the problem down.)

DanielGibson:
Furthermore, I wonder if all text is missing. Can you open the console (Shift+Esc)? Any text there?

tlosm:
Hi Daniel, 
negative the text in the console are show right, only the in game (datapad, terminal etc) and menu are missing

DanielGibson:
(someone else who actually has Big Endian hardware will have to debug this)

dhewg:
[...]

PR #226, opened 2019-02-11
Fix build on powerpc but not mac

AMDmi3:
...by including `vecLib/vecLib.h` only on `__APPLE__`, and not just any `__ppc__`

From https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=235668:

> On powerpc* platforms vecLib/vecLib.h is included by default.
> This header is only available on Mac OS X, so remove it. It
> doesn't seem to be actually needed to compile this port.

### DIFF
--- a/neo/renderer/tr_main.cpp
+++ b/neo/renderer/tr_main.cpp
@@ -26,7 +26,7 @@ If you have questions concerning this license or the applicable additional terms
 ===========================================================================
 */
 
-#ifdef __ppc__
+#if defined(__ppc__) && defined(__APPLE__)
 #include <vecLib/vecLib.h>
 #endif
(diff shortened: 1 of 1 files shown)


pkubaj:
I just wanted to send a PR for that patch, but you were faster :)

DanielGibson:
Thanks a lot! :)

BTW, do you happen to have PPC desktop hardware that can run dhewm3?
If so, I'd like to know if you can reproduce https://github.com/dhewm/dhewm3/issues/148, i.e. if you start the game does the main menu look like it's supposed to or is text missing?
(You don't have to debug it, I just wonder if this is really a general PPC/Big Endian issue, or maybe just a driver bug on the reporters system)

PR #229, opened 2019-02-23
Remove use of date/time macros

### LINKED ISSUE #172: __DATE__ and __TIME__ macros used in source
Use of these macros prevent reproducible builds, which is relevant to Linux distro packaging.
If removing them is not an option, it should be possible to disable them through the build system.

mia-0:
Fixes #172

### DIFF
--- a/neo/d3xp/Game_local.cpp
+++ b/neo/d3xp/Game_local.cpp
@@ -328,7 +328,6 @@ void idGameLocal::Init( void ) {
 
 	Printf( "----- Initializing Game -----\n" );
 	Printf( "gamename: %s\n", GAME_VERSION );
-	Printf( "gamedate: %s\n", __DATE__ );
 
 	// register game specific decl types
 	declManager->RegisterDeclType( "model",				DECL_MODELDEF,		idDeclAllocator<idDeclModelDef> );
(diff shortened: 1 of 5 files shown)


DanielGibson:
Thanks for the patch, sorry I took so long to look at it!

I'm a bit hesitant about merging it, mainly for two reasons:
1. Would be nice if it could be enabled/disabled via CMake
2. com_version/si_version (which uses `version` in Common.cpp) is used in networking code, see https://github.com/dhewm/dhewm3/blob/master/neo/framework/async/AsyncClient.cpp#L1910 (I think that part communicates with id's master server?) - so I'm not sure if changing the strings format would break anything.. (we could still use a dummy date for reproducible builds - or somehow use the date/time of the latest git commit?)
(Not sure if g_version, which uses `gameVersion` from SysCvar.cpp, is used anywhere; if so, it's not obvious at least)

DanielGibson:
I made this configurable via cmake, using hardcoded (in neo/framework/Licensee.h) date/time values for reproducible builds: https://github.com/dhewm/dhewm3/commit/5e7019ebfef7dfe039482ab0d8c3957238b34cf8

PR #242, opened 2019-07-07
Fix ppc64le build

anengineer1:
Special thank to frediz who is responsible for fixing the build in Debian. I just borrowed the patch.


DanielGibson:
Thanks!

anengineer1:
Always a pleasure :-)

PR #254, opened 2019-10-05
Fix 'no specular' support

RalfVB:
Light entities can have the param 'nospecular' set, which skips the specular lighting.
However this never actually worked in Doom 3.
This fix addresses that.

Using IceColdDuke's PBR interaction shader (which exposes the problem clearly), this is without the fix: (nospecular is set on the laser, yet they are still generated, hence the strange lighting around the laser dot)
https://imgur.com/lIst0NA

With the fix applied (specular lights are properly skipped)
https://imgur.com/qO0m34V

### DIFF
--- a/neo/renderer/tr_render.cpp
+++ b/neo/renderer/tr_render.cpp
@@ -682,160 +682,164 @@ static void RB_SubmittInteraction( drawInteraction_t *din, void (*DrawInteractio
 /*
 =============
 RB_CreateSingleDrawInteractions
-
 This can be used by different draw_* backends to decompose a complex light / surface
 interaction into primitive interactions
 =============
 */
 void RB_CreateSingleDrawInteractions( const drawSurf_t *surf, void (*DrawInteraction)(const drawInteraction_t *) ) {
-	const idMaterial	*surfaceShader = surf->material;
-	const float			*surfaceRegs = surf->shaderRegisters;
-	const viewLight_t	*vLight = backEnd.vLight;
-	const idMaterial	*lightShader = vLight->lightShader;
-	const float			*lightRegs = vLight->shaderRegisters;
-	drawInteraction_t	inter;
-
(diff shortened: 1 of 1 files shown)


DanielGibson:
Somehow all the tabs used for indentation have been replaced with spaces, which makes it hard to tell what has actually changed in that function (vs what is just indented differently). It also makes the code look inconsistent in the editor and `git blame` will also be less useful for the affected code..

Can you try keeping the original indentation style?

DanielGibson:
Also, doesn't this change the look of the original game?

RalfVB:
Ahh, the tabs > spacing thing might be related to my text editor. 

Afaik it doesn't affect the base game at all. If they actually used nospecular anywhere, they were bound to notice it didn't work, right?

DanielGibson:
There's 30 maps that contain `"nospecular" "1"`, 1043 occurrences.
I checked and it's indeed set at light entities.
All those lights will look different when this behavior is fixed..

RalfVB:
Then dismiss the patch.

DanielGibson:
Maybe there's a way to make sure it's only applied for new maps (where we can assume they have been tested with the fix applied and thus look as intended)?
Like a `"usenospecular" "1"`flag in the worldspawn?
Do you think that'd be useful for mappers/modders?

RalfVB:
That could be a good solution, yes

RalfVB:
A better approach might be to control it via a cvar?

DanielGibson:
Why do you think that would be better?
I thought it would be best to let mappers specify that so the user doesn't have to change a cvar between playing old and new maps

Arl90:
The idea of a worldpsawn flag is great. I would also add a cvar (disabled by default) to enable the correct behavior of this feature in the entire game.

DanielGibson:
@Arl90 @EoceneMiacid Is anyone still interested in this? 

If so, see https://github.com/dhewm/dhewm3/pull/643
Please test that and provide a testmap, else I won't merge it!

PR #282, opened 2020-03-31
Add CVar to allow the enabling of always run in singleplayer

### LINKED ISSUE #278: Author's opinion on enabling alwaysrun in SP?
Howdy.

What is the author's stance on allowing alwaysRun to be enabled in singleplayer?
It is not vanilla behaviour, however personally I don't foresee it negatively affect any Doom 3 player no matter their stance on the subject.

Those that want it, get it. Those that don't want it, won't have to change a thing as it's not the default behaviour.

My current implementation just rips out the multiplayer check for that specific feature. It is still controlled by the cvar and the toggle in the multiplayer menu.
I assume the main thing that would have to be addressed here, is the fact that the toggle from the multiplayer menu would now affect the game in singleplayer.

Adding another cvar just to control the same feature in sp seems superfluous to me. What would be a good approach, assuming this feature is accepted by the author?

dobosken:
Resolves #278.
Tested in singleplayer, and made sure it doesn't affect the game during multiplayer.

### DIFF
--- a/neo/framework/UsercmdGen.cpp
+++ b/neo/framework/UsercmdGen.cpp
@@ -391,6 +391,7 @@ private:
 	static idCVar	in_angleSpeedKey;
 	static idCVar	in_freeLook;
 	static idCVar	in_alwaysRun;
+	static idCVar	in_alwaysRunSP;
 	static idCVar	in_toggleRun;
 	static idCVar	in_toggleCrouch;
 	static idCVar	in_toggleZoom;
@@ -408,6 +409,7 @@ idCVar idUsercmdGenLocal::in_pitchSpeed( "in_pitchspeed", "140", CVAR_SYSTEM | C
 idCVar idUsercmdGenLocal::in_angleSpeedKey( "in_anglespeedkey", "1.5", CVAR_SYSTEM | CVAR_ARCHIVE | CVAR_FLOAT, "angle change scale when holding down _speed button" );
 idCVar idUsercmdGenLocal::in_freeLook( "in_freeLook", "1", CVAR_SYSTEM | CVAR_ARCHIVE | CVAR_BOOL, "look around with mouse (reverse _mlook button)" );
 idCVar idUsercmdGenLocal::in_alwaysRun( "in_alwaysRun", "0", CVAR_SYSTEM | CVAR_ARCHIVE | CVAR_BOOL, "always run (reverse _speed button) - only in MP" );
+idCVar idUsercmdGenLocal::in_alwaysRunSP("in_alwaysRunSP", "0", CVAR_SYSTEM | CVAR_ARCHIVE | CVAR_BOOL, "always run (reverse _speed button) - only in SP");
 idCVar idUsercmdGenLocal::in_toggleRun( "in_toggleRun", "0", CVAR_SYSTEM | CVAR_ARCHIVE | CVAR_BOOL, "pressing _speed button toggles run on/off - only in MP" );
 idCVar idUsercmdGenLocal::in_toggleCrouch( "in_toggleCrouch", "0", CVAR_SYSTEM | CVAR_ARCHIVE | CVAR_BOOL, "pressing _movedown button toggles player crouching/standing" );
 idCVar idUsercmdGenLocal::in_toggleZoom( "in_toggleZoom", "0", CVAR_SYSTEM | CVAR_ARCHIVE | CVAR_BOOL, "pressing _zoom button toggles zoom on/off" );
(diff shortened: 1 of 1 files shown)


DanielGibson:
I haven't tested this (and, like in your other PR don't have much time for it right now), but is this really enough? Wouldn't you also have to disable slowing down because you ran out of stamina?
Or am I just missing something?

dobosken:
No, that's intentional. _in_alwaysRun_ (and thus the new counterpart _in_alwaysRunSP_) only dictates that the functionality of the sprint key is reversed (walk becomes run and vice versa).
If players want to actually run forever, they can set pm_stamina to 0.

Confusing yes, but it adheres to id Software's CVar naming scheme.

The only behaviour I currently have my doubts about, is that stamina does not regenerate once the bar has run out and the player reverts to a walking pace again. It will only regenerate if the player holds the sprint key and initiates walking manually.


We can't have it regain stamina, 'cause that would just immediately initiate sprinting again, thus depleting the little stamina that was regenerated, causing some weird oscillating behaviour between sprinting and walking.

An alternative would be to follow a sort of modern game design trend, wherein stamina does regenerate, but sprinting is disabled until it is partially filled again (20%? 30%? 40%?).
As there is no precedent for this from id Software in vanilla Doom 3, we would have to define such behaviour ourselves.

Personally I am not comfortable with that, and prefer the current behaviour, as awkward as it may be.

DanielGibson:
Hmm but isn't always run kinda pointless if you can only run for a few seconds and then have to press the "run" key to walk so you can run again?

OTOH, maybe it's enough to document (maybe also in the CVars description) that one probably wants to set pm_stamina to 0 if using this..

dobosken:
You're right, pm_stamina 0 is a great counterpart to this new CVar and pretty much essential.

I think external documentation (websites, changelogs, forums etc.) will suffice. in_alwaysRunSP is an entirely new CVar not documented anywhere else yet, so it'll have to be written down somewhere anyway. Might as well add the pm_stamina 0 tidbit with it.

Perhaps a _dhewm3-exclusive CVars_ section for the website? I'm not sure how many there are right now, but assuming my other pull request is merged, that'll make at least 2 😛

thoughtartist:
Could there be a way to inverse the run and walk speeds themselves in a CVar without modding the player.def file?

DanielGibson:
I tested this a bit and came to the conclusion that this **really** should set pm_stamina to 0 and make sure it doesn't get overriden - neither by `idTarget_EnableStamina`, nor by `idPlayer::Restore()`, nor in `idPlayer::Init()` where all the "pm_*" spawnflags are set as cvars.
Setting `pm_stamina 0` once would be ok, but it seems to get reset frequently when playing (on every level load at least?) and that's just annoying.

dobosken:
Ah, I didn't know _pm_stamina_ resets. I tested it with a modified player.def, which survives the map transitions. My apologies for not being more thorough.

Yeah, that does hamper out-of-the-box usability quite a bit.

I can make it bypass _pm_stamina_ and just outright disable stamina drain from the run routine if _in_alwaysRunSP_ is set to 1?
Doing this kills player autonomy and prevents them from adjusting/balancing stamina values with alwaysRun enabled entirely (they might want alwaysRun gameplay with stamina management, just tweaked from the default game), but it's a bit wacky in how it behaves anyway due to the whole _doesn't regen unless the player manually initiates a walk again_ thing.

Any other options would be too "dirty" I think :/

DanielGibson:
> My apologies for not being more thorough.

No worries, that's why I test :)

> Doing this kills player autonomy and prevents them from adjusting/balancing stamina values with alwaysRun enabled entirely (they might want alwaysRun gameplay with stamina management, just tweaked from the default game), but it's a bit wacky in how it behaves anyway due to the whole doesn't regen unless the player manually initiates a walk again thing.

I don't think there's much value in using always run when you run out of stamina all the time, so I think it's ok to either bypass `pm_stamina` or set it to 0 automatically (and make sure it doesn't get reset as long as `in_alwaysRun` is set to 1) - at least I think that's less surprising to the player (and more useful) than the alternative.

Bypassing pm_stamina *might* have the advantage that the stamina bar in the HUD doesn't get into hell-mode (constant scrolling animation) which might be distracting? But IDK if that's really an issue.

I don't care much either way about bypassing vs setting to 0, so I'll leave that up to you.

One minor nitpick: Would be helpful to mention the cvar name in the commit message (ideally in the commit message's title) so one can quickly look it up without going through the code changes :)
(Yes, I'll also try to remember to document the dhewm3-specific cvars for endusers somewhere)

Thanks for working on this!

DanielGibson:
I just had another idea regarding running: What about having a "toggle run" action instead?
So if you bind it, and press that button, you'll run until you press it again - that would be useful both with stamina enabled and disabled.

Would be nice to have a CVar to disable stamina (that's more persistent than pm_stamina) in addition, though.

paulwratt:
this seems like the most practical way to achieve "always run" in single player. if there is an action AND CVar, then a server can also control it (if wanted).

DanielGibson:
In https://github.com/dhewm/dhewm3/pull/576 I'm adding a new CVar `in_allowAlwaysRunInSP`
When that's set, in_alwaysRun and in_toggleRun also work in SP.

Haven't done anything about the stamina (yet?) though

PR #283, opened 2020-04-07
Use SDL2's own find_package, instead of the shipped one.

coldtobi:
The shipped one has some issues e.g with multiarch support and
breaks there. Newer SDL2 (>= 2.0.4) ships a cmake file which
can be used. 

(So the downside is that minimum SDL2 version requirement needs to be bumped >= 2.0.4.)


DanielGibson:
hmm unsure how/if that'd work on Windows or macOS..
I'd have to try,  but not before the upcoming 1.5.1 release (RC1 was released today).

BTW, dhewm3 now has a reproducible build option in CMake, might be relevant for packaging :)

HarrievG:
I think I have an issue here too , on windows. I did not have any SDL linking problems when i compiled rbdoom, i think this can go in

DanielGibson:
I merged this (manually), thanks for the PR!

(As expected it broke the Windows build, but that's SDL2's fault, they're shipping a broken sdl2-config.cmake with their -dev packages for Windows)

parkerlreed:
Seems this may be affecting Linux as well

#340

DanielGibson:
This seems to break the build both on Windows when installing the dependencies with vcpkg and on Linux with whatever @parkerlreed uses (please clarify).
I should probably revert this change.. or we need to figure out how to use SDL2's integrated ` sdl2-config.cmake` *properly*, assuming we're doing something wrong right now - even though it looks right to me and works on my system (XUbuntu 18.04) and even on Windows with dhewm3-libs (after I adjusted sdl2-config.cmake of dhewm3-libs)..

parkerlreed:
Yeah Arch based for me right now exhibits the beahvior (Both regular x86_64 and AArch64)

DanielGibson:
I reverted this (for now?).

If there's a solution that works with both sdl2-config.cmake and SDL2Config.cmake (like they are shipped in relevant versions of SDL2, I think at least sdl2-config.cmake has changed a bit between version 2.0.8 in the still relevant Ubuntu 18.04 and the current 2.0.14) I'd be willing to try getting rid of FindSDL2.cmake again.

PR #285, opened 2020-05-12
Allow users to set a custom master server if they want

Stradex:


DanielGibson:
We talked about this on twitter and I was wondering whether ids master server is still running and working as it should.
I think it is.
Thing is, dhewm3 has a different `ASYNC_PROTOCOL_VERSION` than Doom3 1.3.1 (`ASYNC_PROTOCOL_MINOR` is 42 instead of 41, see https://github.com/dhewm/dhewm3/blob/master/neo/framework/BuildDefines.h#L98-L109) so the (few) existing Doom3 servers on the net aren't shown - and probably there are no servers running dhewm3?

Also, it looks like `net_master1` etc indeed aren't really used, except for sending heartbeats to.
`idAsyncNetwork::masters[]` is an array with 5 elements, initialized to the 5 net_master* cvars, but only accessed via `idAsyncNetwork::GetMasterAddress()` (which always uses index 0) and `idAsyncNetwork::GetMasterAddress( int index, netadr_t &adr )` which is almost always called with index 0, except `idAsyncServer::MasterHeartbeat()` which apparently sends a heartbeat to all the master servers.
This means that `net_master1` indeed isn't used when searching for multiplayer servers.

I'm not sure we really need alternative master servers if the original one is still running and apparently working (though I didn't try running a server myself to see if it's properly registered there and then turns up in the list).

Stradex:
I already tried hosting servers using dhewm3 and you can't see the server in the server lists. Probably there is some hard-coded thing at id master servers about ASYNC_PROTOCOL_VERSION or who knows.

DanielGibson:
Were those servers reachable from the internet via IPv4?

Stradex:
Yes, I played a lot of games with friends using connect ipv4:port command via console, because servers are never shown at the server browser. But honestly if we can't use the original master server from ID, it would be nice to have an official dhewm3 master server, instead of allowing users to actually change the master server.

Stradex:
> did they advertise the right IP to the master server?

The heartbeat is always sent to net_master0 address, so I am sure the master server from id receives the heartbeat, but maybe there is a filter or who knows. I am not sure about that. Would be nice to actually figure out that first talking with someone from ID If that is still possible.

DanielGibson:
> But honestly if we can't use the original master server from ID, it would be nice to have an official dhewm3 master server, instead of allowing users to actually change the master server.

I agree.
I'm still curious why the id master server apparently doesn't work though, maybe it's just a little bug in dhewm3 that could be resolved.
(Making it work might have the advantage that existing dhewm3 versions would also work, assuming that at least the client side of it isn't broken)

Stradex:
GPL version of Doom 3 actually works with master servers I think so, so it would be nice to find what is causing the problem. I can test trying to set back  ASYNC_PROTOCOL_VERSION to the original doom3 1.3.1 just for testing purposes in my local machine, and see if that is actually the problem or not.

DanielGibson:
Good idea!

Stradex:
Fucking lol, I got it working after forcing: si_version to DOOM 1.3.1.1304 win-x86 Jan 16 2007 15:41:40 in common.cpp

So as I guessed, it is hardcoded in the master server

DanielGibson:
oh FFS

Does it have to be exactly that or is it enough if it starts with "DOOM" or whatever?

Stradex:
> oh FFS
> 
> Does it have to be exactly that or is it enough if it starts with "DOOM" or whatever?

I will test a bit, but in fact we can "inject" the code avoiding having to touch si_version

DanielGibson:
I mean, if we're lucky it's like a browser user agent string and we only have to start with the right string (that hopefully is not all of what you pasted there) and can add the real dhewm3 version to the string behind that

PR #286, opened 2020-05-16
Fix display on macOS with Retina display

### LINKED ISSUE #276: Display anomolies on Mac
I tried building off both master and 1.5.0 release.

On Mac OS Catalina 10.15.3.

Dhewm displays the game only in the bottom 1/4 of the open window, in windowed mode.

In Full Screen mode, the entire display is shifted 1/4 up, so the top 1/4 of the game is not visible.

ioan-chera:
This patch fixes the wrong displaying on macOS with Retina display on the lower-left corner. The fix is explained here: https://stackoverflow.com/a/58666167/11738219

This shall fix #276 .

I also added a recommended setting in the .plist, though I'm not sure how relevant it is for this: NSHighResolutionCapable.

### DIFF
--- a/neo/sys/glimp.cpp
+++ b/neo/sys/glimp.cpp
@@ -95,7 +95,7 @@ bool GLimp_Init(glimpParms_t parms) {
 
 	assert(SDL_WasInit(SDL_INIT_VIDEO));
 
-	Uint32 flags = SDL_WINDOW_OPENGL;
+	Uint32 flags = SDL_WINDOW_OPENGL|SDL_WINDOW_ALLOW_HIGHDPI;
 
 	if (parms.fullScreen)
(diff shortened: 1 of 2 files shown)


ioan-chera:
Closing this too, as it's still broken if I try different resolutions while in fullscreen. I'll reopen this when it becomes robust.

DanielGibson:
When you've got it working, can you also test the mouse cursor speed in windowed mode?
The games cursor (e.g. in the main menu) *should* be about as fast as the system cursor, but I can imagine that HighDPI can screw that up (if the pixel coordinates from the mouse events are lowdpi)

DanielGibson:
Forgot to mention, a good way to compare the cursors is running in windowed mode and opening the console (Shift+Esc), then you might be able to see both the system cursor (possibly outside the window) and the ingame one.

PR #287, opened 2020-05-16
Add fullscreen-desktop video mode

ioan-chera:
Add a new value to r_fullscreen, numbered 2, which can be used to set SDL mode as SDL_WINDOW_FULLSCREEN_DESKTOP, as an alternative to the classic SDL_WINDOW_FULLSCREEN. It has the benefit of not changing user's monitor mode and making it easier to switch out of the app (notably on macOS), while the drawback is less processing priority on the game if other stuff are happening.

I understand that the game GUI is not going to access this, so this will have to be documented from the config settings.


ioan-chera:
Actually hold this off… while it fixes the issue, it doesn't really do the intended job of reducing resolution on fullscreen.

PR #293, opened 2020-05-27
Master server working again

Stradex:
So this was the real issue after all. (the thing about si_version etc.. was nothing but a false-positive because I was testing after using the original Doom 3 steam version so probably the cache did saved my ip and port in the masterserver for a while)

### DIFF
--- a/neo/framework/async/AsyncServer.cpp
+++ b/neo/framework/async/AsyncServer.cpp
@@ -2069,6 +2069,7 @@ void idAsyncServer::ProcessGetInfoMessage( const netadr_t from, const idBitMsg &
 		outMsg.WriteString( sessLocal.mapSpawnData.userInfo[i].GetString( "ui_name", "Player" ) );
 	}
 	outMsg.WriteByte( MAX_ASYNC_CLIENTS );
+	outMsg.WriteInt( -1 );
 
 	serverPort.SendPacket( from, outMsg.GetData(), outMsg.GetSize() );
 }


Stradex:
A bit more of explanation why this works:

dhewm3 code did remove the following line:

outMsg.WriteLong( fileSystem->GetOSMask() );

because GetOSMask does not exists anymore, but the master server expects that info to be received seems (even if there is no use for them maybe), so adding outMsg.WriteInt( -1 ); fixsthe problem.

turol:
You should add a comment about what this magic `-1` is and why it is.

Kalamatee:
it would be nice if it was set using a cvar that could be overridden if
necessary.

DanielGibson:
> it would be nice if it was set using a cvar that could be overridden if necessary.

when should that be necessary?

Stradex:
Can someone do the pull request with the comment for me?, I am to lazy for do a pull request again for just adding a comment lol.

DanielGibson:
I can add the comment, of course :)

Note that you wouldn't have to create a new PR - you could just add a commit to the branch and push it (it gets added to the PR then) or even amend the existing commit and force-push the branch.

Stradex:
Yeah but I need to restart the pc (I am in linux now, I have everything in windows), open gitbash and aah, I am getting lazy to just thing about it right now. haha

btw, the feature was already tested and it is working. Just to you know that this is not a feature without testing.

DanielGibson:
I pushed the commit to master (after adding a comment), thanks a lot! :)

PR #296, opened 2020-06-01
Update sounds at ~60Hz instead of ~10Hz, fixes #141

### LINKED ISSUE #141: Sound buffers?
One thing has always bothered me about id Tech 4 games - Doom 3, Quake 4, and Prey specifically; Wolfenstein doesn't seem to have this problem - and that's the way rapid-fire weapons sound. Under sustained fire, the machine gun, chaingun, and plasma rifle all sound as if they're "missing" their firing sounds every few shots. Other things suffer from this, too, like footstep noises. I suspect it's a problem with the sound buffer size and response times, but I can't find any console variables to set to fix it.

I'd asked about this on the Doomworld forums, but was directed to ask here instead; is there some oddity to the id Tech 4 sound engine at work here? I can confirm that it happens in both vanilla Doom 3 (Windows 1.3.1) and in a release build of dhewm3 from last week.

DanielGibson:
For some reason sounds were only updated/started about every 100ms, which could lead to delays of up to around 100ms after a sound gets started (with `idSoundEmitterLocal::StartSound()`) until OpenAL is finally told to play the sound.
Also, the actual delay drifted over time between 1ms and 100ms, as the sound ticks weren't a fixed multiple of the (16ms) gameticks - and the sound updates didn't even happen at the regular 92-94ms intervals they should because they run in the async thread which only updates every
16ms...
Because of this, the machine gun and other rapid firing weapons sounded like they shot in bursts or left out shots occasionally, even though they don't.
Anyway, now sound is updated every 16ms in the async thread so delays are <= 16ms and hopefully a lot less noticeable.

https://github.com/dhewm/dhewm3/issues/141#issuecomment-636620498 has more details on this.

You can set `com_asyncSound 2` to get the old behavior for comparison (the meaning of com_asyncSounds values have changed a bit with this change), `com_asyncSound 1` (or 3) give you the new behavior.

**Windows binaries with this fix for testing**: [dhewm3-oal-timing-fix.zip](https://github.com/dhewm/dhewm3/files/4713493/dhewm3-oal-timing-fix.zip)

I tested it a bit an didn't notice any issues, but it definitely **needs thorough testing** for any new sound-related issues *anywhere* in the game before going into master - thanks in advance for any feedback!

### DIFF
--- a/neo/framework/Common.cpp
+++ b/neo/framework/Common.cpp
@@ -83,12 +83,8 @@ idCVar com_purgeAll( "com_purgeAll", "0", CVAR_BOOL | CVAR_ARCHIVE | CVAR_SYSTEM
 idCVar com_memoryMarker( "com_memoryMarker", "-1", CVAR_INTEGER | CVAR_SYSTEM | CVAR_INIT, "used as a marker for memory stats" );
 idCVar com_preciseTic( "com_preciseTic", "1", CVAR_BOOL|CVAR_SYSTEM, "run one game tick every async thread update" );
 idCVar com_asyncInput( "com_asyncInput", "0", CVAR_BOOL|CVAR_SYSTEM, "sample input from the async thread" );
-#define ASYNCSOUND_INFO "0: mix sound inline, 1: memory mapped async mix, 2: callback mixing, 3: write async mix"
-#if defined( __unix__ ) && !defined( MACOS_X )
-idCVar com_asyncSound( "com_asyncSound", "3", CVAR_INTEGER|CVAR_SYSTEM|CVAR_ROM, ASYNCSOUND_INFO );
-#else
-idCVar com_asyncSound( "com_asyncSound", "1", CVAR_INTEGER|CVAR_SYSTEM, ASYNCSOUND_INFO, 0, 1 );
-#endif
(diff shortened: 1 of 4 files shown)


BenzoSt:
Gave the weapons a try in test_box, and all is working fine for me.  A definite improvement.

thedrakeraider:
I know this is a crazy bump, and I just found out about this researching this exact issue, but do you think this fix can be piped back into the vanilla exe via a patch of some kind? One of my biggest issues with the game.

shuckster:
@thedrakeraider You can *replace* the vanilla EXE with the [current build of dhewm3](https://github.com/dhewm/dhewm3/releases/tag/1.5.2).

Or just pop it alongside, giving you the choice between running the fixed version and the OG.

DanielGibson:
it should be possible to port this fix to the original doom3 gpl source, for example to build a doom3.exe that is compatible with old mods.
note that you may need more dhewm3 commits than the ones from this pull request, as far as I remember I later did more sound fixes that might have been needed due to this change (sth about endless loops? I forgot the details)

DanielGibson:
however it's possible that this fix only works with openal sound (original doom3 also had a software mixing sound backend, and I think there openal was only supported on windows, dhewm3 only supports openal for sound, on all platforms)

Issue #261, opened 2019-11-03
com_fixedtic -1 causes Imp fireball and Lost Soul trail effect to render incorrectly


shoober420:
When using com_fixedtic -1, the Imp fireball smoke effect and Lost Soul smoke trail when it flys toward you renders incorrectly. It causes weird graphical corruption with shadows in the air. Very odd. I’ll post my specs below.

OS : Arch Linux 64bit (Testing Repository)

--- Kernel : 5.3.8-zen+
--- File System : ext4
--- Shell : zsh
--- Display Server : X.Org 1.20.5
--- Window Manager : bspwm
--- Terminal Emulator : st-scrollback (fg: #00ff7f) (bg: #000000) 
--- Font : IBM Plex Mono
--- File Manager : cd, cp, rm, ls, ln
--- Text Editor : vim
--- Task Manager : htop
--- Audio Player : cmus
--- Video Player : mpv
--- Image Viewer : sxiv
--- Video Editor : avidemux
--- Raster Graphics Editor : azpainter
--- IRC Client : weechat
--- Disc Burner : libburn
--- Web Browser : Chrome Dev
--- Steam Client Beta 

Case : Cooler Master Elite 310 (Noctua Fans) 
Motherboard : ASUS P5QL
PSU : Corsair CX750M
CPU : Intel Core2Quad Q9650 @3.8GHz
GPU : NVIDIA GeForce GTX 760 2GB (440.26)
SPU : ASUS Xonar Essence STX
RAM : 8GBs DDR2 800MHz
NIC : Intel Gigabit CT Desktop Adapter
Mouse : Microsoft Intellimouse Explorer 3.0 @500Hz (Zowie P-SR) 
Monitor : IBM P275 CRT
Keyboard : SteelSeries 6Gv2 PS/2 (Cherry MX Red) 
Headphones : Grado Prestige SR80e

shoober420:
Carmacks reverse and StencilOpSeparate, whether they are on or off, has no effect. The issue still remains.

DanielGibson:
Yeah, I think the trick is to set com_fixedTic back to 0 :-P

At what refreshrate did you play?

https://github.com/dhewm/dhewm3/issues/248#issuecomment-530482578 also mentions problems (flickering particles; maybe the same as yours?) with com_fixedTic -1

[...]

Issue #248, opened 2019-08-25
issues with Dentonmod; resolving them will likely also allow me to submit my mod for official mod support


BenzoSt:
I recently replaced my Nvidia video card with an AMD Radeon video card. This caused problems with running several standard Doom mods (with Doom3.exe, not Dhewm3.exe), including my own custom mod, Infernal Death Orgy, which was in turn based on Denton's Enhanced Mod 2.02.

Dhewm3 has official support for Denton's Enhanced Mod 2.02, and it indeed works properly, whereas Denton's mod won't work with standard Doom3.exe when I am using my AMD Radeon video card.

But there is a glaring issue with both Denton's mod and my own mod when using Dhewm3 with my AMD Radeon video card: Several of the particle effects flicker. Try running Denton's Mod with  Dhewm3 and a Radeon card and fire the rocket launcher a few times and you will see.

Side note: Also, is there a way to disable the echoing sound effects when playing with Dhewm3?

DanielGibson:
Regarding the echoing sound effects: disable EAX in the sound settings

Regarding dentonmod, I'll have to look at that, I hope I can reproduce it on my windows+radeon PC, thanks for reporting!

BenzoSt:
I noticed that setting com_fixedtic to 0 in the console fixed the problem with the flickering particles.

DanielGibson:
what was com_fixedtic set to before?

BenzoSt:
Com_fixedtic was previously set to “-1”. When I set it to “0”, the flickering particles problem was corrected.

Sent from Mail for Windows 10

DanielGibson:
Ok; com_fixedtic -1 is still experimental and has some known issues (I think #230 discusses this a bit; https://github.com/dhewm/dhewm3/issues/15#issuecomment-521099340 mentions that he has some fixes for timing problems).
So I guess your issue should be added to the (right now imaginary) list of things to fix when adding proper support for >60Hz.

(I only have 60Hz displays so I can't properly work on that myself :-/)

DanielGibson:
closing as it was broken due to com_fixedtic -1

Issue #230, opened 2019-03-03
Smooth movement on refresh rates above 60Hz?


dezo2:
Recently I found a simple tool named Quake 4 Tweaker, which allows Q4 to run smoothly on monitors with refresh rate above 60.
https://community.pcgamingwiki.com/files/file/1009-quake-4-tweaker/
I think it does that by hacking into the global game timing slowing it down to the selected monitor refresh rate / target FPS. The more target FPS is selected, the slower the game movement/animation is, so it simply compensates for the game speed up above 60Hz when using uncapped framerate (com_fixedTic 1). The result is perfectly smooth gameplay with correct timing on higher refresh rates. Is it possible to implement something like this into the dhewm3 code? I tried the "experimental" com_fixedTic -1 but for me it behaves the same as com_fixedTic 0, so on the higher refresh rates (100, 120, 144) the movement is just a stuttery mess capped to 63Hz.

dezo2:
After some source code digging I think that all it takes is a little change in UsercmdGen.h:
  const int USERCMD_HZ = 120;  //60
Compiling the source code will produce new base.dll - replace it in the dhewm3 installation and then use autoexec.cfg with:
  seta com_fixedTic "1"
  seta r_swapInterval "1"
  seta r_finish "1"  ;optional to reduce input lag
That should lock dhewm3 movement to smooth 120FPS just like in bfg edition. Cutscenes timing will be still al little off (that would need multiples of 62.5Hz refresh rate), but thats a minor problem compared to 60FPS judder on 120Hz refresh rate. I will test it some more, but so far it seems to be working OK.

KillPixelGames:
This would be amazing.  Playing Doom3 at 60hz is brutal.  IMO, the 120hz of the BFG edition is its only saving grace.

I really hope this gets implemented in dhewm3 :D

buvk:
Could you share the dll? I'm having issues compiling the source in mysys.  Is it possible to go higher? E.g. 240hz

dezo2:
I am not that familiar with the sources, but I think Daniel should be able to create a cvar for this. Even if it should require restart of the engine, it will be better than hardcoded value. I didn't play with this since march, but I can share the 120Hz dll here for version 1.5.0, if you want to try it. It should work with any other frequency, but I tried only 120Hz as I have currently 120Hz monitor. Just make sure your autoexec.cfg is set as above. The dll is compiled using VS2017 and latest stable version 1.5.0 of sources.

[...]

Issue #169, opened 2017-04-04
Race conditions between main thread and SDL async timer thread


turol:
ThreadSanitizer report:
```
WARNING: ThreadSanitizer: data race (pid=25333)
  Read of size 4 at 0x000001ddbdf0 by main thread:
    #0 idSessionLocal::DrawWipeModel()                            neo/framework/Session.cpp:2220   (dhewm3+0x00000066b8e9)
    #1 idSessionLocal::Draw()                                     neo/framework/Session.cpp:2454   (dhewm3+0x00000066b8e9)
    #2 idSessionLocal::UpdateScreen(bool)                         neo/framework/Session.cpp:2497   (dhewm3+0x0000006742e9)
    #3 idSessionLocal::StartPlayingRenderDemo(idStr)              neo/framework/Session.cpp:937    (dhewm3+0x00000066ca4a)
    #4 idSessionLocal::TimeRenderDemo(char const*, bool)          neo/framework/Session.cpp:965    (dhewm3+0x00000066d178)
    #5 Session_TimeDemoQuit_f(idCmdArgs const&)                   neo/framework/Session.cpp:681    (dhewm3+0x000000676907)
    #6 idCmdSystemLocal::ExecuteTokenizedString(idCmdArgs const&) neo/framework/CmdSystem.cpp:498  (dhewm3+0x0000005f23d5)
    #7 idCmdSystemLocal::ExecuteCommandBuffer()                   neo/framework/CmdSystem.cpp:679  (dhewm3+0x0000005f2b0b)
    #8 idEventLoop::RunEventLoop(bool)                            neo/framework/EventLoop.cpp:184  (dhewm3+0x0000006442b7)
    #9 idCommonLocal::Frame()                                     neo/framework/Common.cpp:2384    (dhewm3+0x0000006004f0)
    #10 main                                                      neo/sys/linux/main.cpp:305       (dhewm3+0x0000007fdff1)

  Previous write of size 4 at 0x000001ddbdf0 by thread T1 (mutexes: write M104):
    #0 idCommonLocal::SingleAsyncTic()  neo/framework/Common.cpp:2497  (dhewm3+0x00000060096f)
    #1 idCommonLocal::Async()           neo/framework/Common.cpp:2540  (dhewm3+0x000000600ab4)
    #2 AsyncTimer(unsigned int, void*)  neo/framework/Common.cpp:2728  (dhewm3+0x00000060181f)
[...]

PR #297, opened 2020-06-11
Add com_gameHz to play with higher HZ/FPS

Stradex:
This basically allows the player to set the amount of HZ/FPS they want the game to run. Making it a bit like BFG Edition in that aspect.

[pak009.zip](https://github.com/dhewm/dhewm3/files/4763228/pak009.zip)


Stradex:
Also, if someone have a mod with custom game logic for dhewm3 and it is worried  that after this pull request that mod is not going to work anymore. This is the necessary code I used to update librecoop to make it work with this pull request. (Also keeping compatibility with original dhewm3 usign a preprocessor def by now)

https://github.com/Stradex/librecoop/commit/7f87c0b3103f2b1e2ebf9d247616c07f65c77698

DanielGibson:
Thanks for the PR, it will most probably take a while till I merge it because, as you said, it might break existing mods.
What seems especially problematic in that regard is that it needs changes to the scripts - so far I could avoid shipping *any* game data, all the dhewm3 changes have been done in code. That wouldn't necessarily be a deal breaker for the main game and RoE, but for mod-compatibility it's more difficult :-/

Also I'll have to review (and ideally test) this thoroughly, and I'm not sure when I have enough time for that.. (unfortunately, I can't even test it properly because all my current screens are 60Hz)

Don't get me wrong, this looks great on first glance and it's a often-requested feature, but as it's a quite invasive change I want to make sure we get it right and minimize breakage :)

Stradex:
Hello!

No problem!, I understand that this can be a huge change and can break compatibility with older mods. And that is a problem yes, I will try to find a way to do some kind of "hack" in order to avoid needing to modify the scripts and add a new .pk4 file. I suppose that if I achieve that then we can have speed up the merge of this feature :)

Arl90:
This seems great, hope we can have it in dhewm 3.

DanielGibson:
> For people with screens that doesn't support more than 60hz, use r_swapInterval 0 (need to close and open the game again after doing it)

entering `vid_restart` in the console might suffice :)

I just remembered that there already was some discussion about >60fps at https://github.com/dhewm/dhewm3/issues/230 and also https://github.com/dhewm/dhewm3/issues/15#issuecomment-521099340 - revelator says he has a timing-related fix from DarkMod, so it might be worth checking out what they did (unfortunately, his modified dhewm3 doesn't seem to be on Github).

In the discussions there was something about Doom3 not properly running at 60fps, but 62.5, and it seemed that this was due to time being measured in integer(!) milliseconds.
```c++
const int USERCMD_HZ			= 60;			// 60 frames per second
const int USERCMD_MSEC			= 1000 / USERCMD_HZ; // 16
```
A frame should be 16.666.. ms, but the code uses 16 everywhere - and 1000/16 = 62.5 (instead of the desired 60) - that (combined with vsync) might lead to micro stuttering, I guess?
And millisecond resolution might cause even more problems at higher framerates (=> lower frametimes, just 4,166.. ms at 240fps).

I wonder if your changes here are affected by these problems?
Might be worth checking out what darkmod and D3BFG do?

I'm not sure if that makes sense or is even possible in Doom3, but do you think we could continue to run the game logic at 60Hz, so scripts and (most of?) the code in the game DLL don't have to be changed, but render at a different rate (and probably update at least the camera so looking around feels smooth)? It *sounds* like it should work at multiples of 60Hz (like 120Hz), but what if someone tries running the game at 90Hz or 144Hz or 165Hz - will stuttering get even worse then?
If running the game code that way (async), we might be able to modify the rest of the engine to either use **micro**second for time, or maybe keep milliseconds but put them in a float or double for additional precision?
I haven't investigated this topic in Doom3, so this is just ideas :)

Stradex:
I have good news, I added a new commit that lets us use this feature without having to add a new pak file.

Stradex:
> In the discussions there was something about Doom3 not properly running at 60fps, but 62.5, and it seemed that this was due to time being measured in integer(!) milliseconds.
> 
> ```c++
> const int USERCMD_HZ			= 60;			// 60 frames per second
> const int USERCMD_MSEC			= 1000 / USERCMD_HZ; // 16
> ```
> 
> A frame should be 16.666.. ms, but the code uses 16 everywhere - and 1000/16 = 62.5 (instead of the desired 60) - that (combined with vsync) might

I did a new commit based on that problem, I believe with the new framerate being float instead of int the improvement should be important. In my end I notice some improvement but I would prefer more people to test it before coming to conlusions too fast.

Stradex:
Sorry, I made a fucking mess with the pull request without noticing :S

The last commit to use should be **fix: multiplayer fix and more**

d6c73042eb47638858cde6b814efd8179bce61c8

Stradex:
This is the last build of this pull request. (Win32 binaries), This is for anyone who wants to test the feature.

[dhewm3 unlocked fps v3.zip](https://github.com/dhewm/dhewm3/files/4777436/dhewm3.unlocked.fps.v3.zip)

DanielGibson:
I looked at TheDarkMod very briefly and while it seems like they did some timing changes (using microseconds instead of milliseconds in a few places, apparently to get real 60fps instead of 62.5), it seems like it's still limited to 60Hz so it's not super helpful for >60Hz support.

I'll also look at D3BFG and I'll also try to understand how exactly a frame in Doom3 works, esp. what part of it happens in the game DLL.
I still hope that we can somehow run the gamecode (incl. physics) at 60Hz, but render at 120Hz (or 240 or whatever).. but for that to work, camera and movement code and ideally also animations should run at 120Hz (or whatever) - otherwise we'd just render the exact same frame twice (or four times), which would be pointless of course and wouldn't look or feel any "smoother".

(If all else fails we could just run at 60Hz and display 120 in the framecounter :-D)


For testing your current solutions it might make sense check if the stuff that was broken with `com_fixedtic -1` or the experimental builds that just hardcoded ` USERCMD_HZ = 120;`.
See:
* https://github.com/dhewm/dhewm3/issues/261 (with com_fixedTic -1) "Imp fireball smoke effect and Lost Soul smoke trail when it flys toward you renders incorrectly"
  * #248 mentions similar problems (flickering particles) with com_fixedTic -1
* https://github.com/dhewm/dhewm3/issues/230#issuecomment-502461387 "When closing the crane in Alpha Labs Sector 3, the game crashes to console"
  * the same thread happens a few more crashes, might be related though

It probably also makes sense to grep the whole engine and gamecode for "60" and "16", in case hardcoded values (instead of USERCMD_HZ and USERCMD_MSEC) are used somewhere (of course not every "60" or "16" in the code should really be one of those constants, we'd have to check every occurrence)

DanielGibson:
Ok, looks like TDM supports 120Hz after all, you gotta search for "uncapped framerate" or "uncapped fps" apparently.
Looks like they also ran into lots of bugs, like
https://bugs.thedarkmod.com/view.php?id=4514
https://bugs.thedarkmod.com/view.php?id=4696
https://bugs.thedarkmod.com/view.php?id=4924
https://bugs.thedarkmod.com/view_all_bug_page.php?filter=5ee6ef1defb1f

I created a git mirror of their SVN repo in case anyone is interested in looked at single commits mentioned in the bugreports: https://github.com/DanielGibson/thedarkmod-mirror
(each git commit message has a line like `git-svn-id: https://svn.thedarkmod.com/publicsvn/darkmod_src/trunk@8769 49c82d7f-2e2a-0410-a16f-ae8f201b507f` - in this case the SVN revision they'd mention in their bugtracker is `8769`)
NOTE: It's big, >400MB! Most probably because they (used to?) check in large static ffmpeg libs for win32 and win64..

nbohr1more:
Thank you for the mirror! 

If you wanna see the bleeding edge TDM stuff, go here:

https://github.com/fholger/thedarkmod

We're (TDM) mostly out of the woods with regard to high-FPS bugs but we probably could have saved ourselves a lot of hassle by simply modifying the core HZ attributes as our new developer Cabalistic did in his branch prior to joining (same dev who maintains the above repo). 

Duzenko ( the developer who began uncapped FPS work ) was impatient to start because he was trying to solve a different timing issue for low FPS scenes and capped FPS was impeding that work. (Eg. That initial implementation had lots of problems.) We have been fortunate that a few outside (and very skilled) contributors mysteriously arrive to fix some of the physics and game bugs (then return to the shadows).

Stradex:
Thank you all for the talk and info, and also thanks to someone who was testing the build and giving me feedback I added a new commit that should be fix a problem related to game physics running a bit more faster than the original game. Also this build is supposed to remove the original mircostutter problem present in vanilla doom 3.

neo/d3xp/Game_local.cpp
```diff
@@ -604,6 +605,8 @@ void idGameLocal::SaveGame( idFile *f ) {
 	savegame.WriteBool( isMultiplayer );
 	savegame.WriteInt( gameType );
 
+	savegame.WriteFloat(preciseTime);
```
DanielGibson, line 608: this breaks savegame compatibility :-(

neo/d3xp/Game_local.cpp
```diff
@@ -334,6 +334,7 @@ void idGameLocal::Init( void ) {
 	//Update MSEC and gameFps
 	gameFps = cvarSystem->GetCVarInteger("com_gameHz");
 	msec = 1000.0f / cvarSystem->GetCVarFloat("com_gameHz");
+	msec *= 0.96f*0.96f; //HACK to emulate OG D3 msec error, in order to have exactly the same game logic speed
```
DanielGibson, line 337: why `0.96f*0.96f` ?
I get that `16 / 16.666 == 0.96`, but why square it?

neo/d3xp/Game_local.cpp
```diff
@@ -309,6 +310,10 @@ void idGameLocal::Init( void ) {
 	const idDict *dict;
 	idAAS *aas;
 
+	msec = 16.0; //60fps
```
DanielGibson, line 313: please put a space after "//", e.g. `// 60fps` (yes, always, not just here), as it's done in the existing code

Stradex:
Thanks for reviewing the code, but tbh with you there are some physics and game logic strange bugs that still I need to fix and I believe that is mainly because my approach probably is not the best at all. Sadly I must admit when what I am doing is not quite right and this is one of these times.

Hopefully maybe this failed attempt will encourage others to actually try this. I am no saying to close this but I am saying that this is not quite right, not even close, to consider a pull request. There is stuff to fix and probably when I have more free time I will try to look for another approach.

neo/d3xp/Game_local.cpp
Stradex: That was a mistake from me. I thought that I needed to do that twice because common frame was already working at 16.666 hz, but multiplying by 0.96f twice was a mistake.

In fact that product for 0.96f is something I don't like at all. It is some kind of "duct-tape" ugly code. I will look at this feature in some days or weeks in order to see if with a fresh mind I can actually figure out something better.

neo/d3xp/Game_local.cpp
Stradex: Thanks for letting me know this. I will try to find a way to avoid that, in fact I was looking at the BFG Edition code like a lot, and there are many things that I don't like about this pull request after looking at that. So I will try to find a way to improve this or actually try this again but most similar to the BFG Edition approach.

DanielGibson: I *think* the best way to do this is to bump `BUILD_NUMBER` and in `idGameLocal::InitFromSaveGame()` you can check `savegame.GetBuildNumber()` and only if it's the new number you read preciseTime, otherwise you set preciseTime to some sane default value.

DanielGibson:
Hey, sorry to hear it doesn't work as well as you hoped!

Can you tell us what bugs you ran into? I think that'd be super-useful to know in case someone else would like to try to fix your approach, or to make sure that other approaches don't have the same issues - or just for yourself when you get back to this later and maybe forgot about the specific bugs :)
(At least I tend to forget such things after a few months)

I turned this pull request to a draft so it's clear that it's not done yet, but I'd otherwise leave it open as I think this contains good information - even if it turns out that indeed a different approach is needed I imagine that *some* of your changes are still useful - and if it's just to see *what* things in the code need to be changed for this.

I'm sure we'll come up with a working solution *eventually*

Sol1vaN:
> When closing the crane in Alpha Labs Sector 3, the game crashes to console"
> * the same thread mentions a few more crashes, might be related though

Seems that I've founded a way to solve this crash (temporal fix).
I'm don't know if this have relation with this problem, but it works for me.

The id3tech engine have a lot of instability when you play in 144Hz/144FP mode (or another Hz/FPS different to 60Hz/FPS).
So, you need to change the afinity of CPU to the 0 core only.
May be the render engine works better in this way when you play at hi Hz/FPS.

Until now, I've played a lot in 120Hz/120FPS mode and right now I'm in the Alpha Labs Sector 4 with no issues.

I hope this help to contribute a little more to find a better way to reach the unlocked FPS feature.

turol:
If CPU affinity helps it then it could be related to #169.

mirh:
I'm just a random passerby but I'm completely missing on the sense of this PR. 
Monitor frequency should only be a concern for the resolution selection code. 
Everything else in the game should just care for framerate. 

If you want v-sync then (putting aside all the fancy [nuances](https://forums.blurbusters.com/viewtopic.php?t=4213) that has today), that should be a completely separate matter until the user decides they want it.

Yamagi:
The point is that Doom 3 (and thus also dhewm3) assumes a max framerate of 60hz throughout the code and within the scripts. Rendering more frames breaks things badly. There are several approaches to solve that problem, however none of them is easy to implement. This PR is one of them.

Being able to render more frames would be nice, because these days 60hz is  the lower end of the spectrum. Even if we can reach only 120hz, it would be a big improvement. VR probably needs 240hz or even more...

Zhnigo:
> This is the last build of this pull request. (Win32 binaries), This is for anyone who wants to test the feature.
> 
> [dhewm3 unlocked fps v3.zip](https://github.com/dhewm/dhewm3/files/4777436/dhewm3.unlocked.fps.v3.zip)

Hello and thanks for your work! Is there a way to implement this functionality in dhewm 1.5.2? 1.5.1 has severe visual bugs when playing with D3HDP

simonedibilio:
I second Zhnigo's comment, I too was trying to get your patch working on 1.5.2 cause of D3HDPv2 (best graphical overhaul out there imo), but - alas - to no avail..

com_fixedTic -1 doesn't really do much for smoothness like "unlockedFPS" does, unfortunately :')

markanini:
> com_fixedTic -1 doesn't really do much for smoothness like "unlockedFPS" does, unfortunately :')

What's the perceivable difference?

simonedibilio:
> > com_fixedTic -1 doesn't really do much for smoothness like "unlockedFPS" does, unfortunately :')
> 
> What's the perceivable difference?

I think it's a frame pacing issue, 'cause fixedTic -1 **feels** like 60hz (or worse!), no matter what: I tried with the game running unlocked at a thousand FPS, or with it limited at various FPS stages (90, 120, 144, 175, 240 - depending on the set monitor's refresh rate) through nvidia's control panel, or nvidia's inspector (which should be the same..), or RTSS (again, should be the same as the control panel).

.. whereas with Stradex's patch, it feels *exactly* like BFG's 120hz mode: perfectly sync'd, smooth and responsive, like in a competitive FPS.

I recently replayed through the whole game on 1.5.1 with this patch and - apart from the memory leak (?) issue that requires you to restart the game every now and then to avoid it getting locked at 25 FPS and the oxygen getting depleted faster on a couple of outside segments of the game (not even all of them) - it was amazing.

simonedibilio:
I decided to take matters into my own hands and ported Stradex's source into dhewm3 1.5.2 .

To anyone landing here from pcgamingwiki or wherever, just check out my repo: https://github.com/simonedibilio/dhewm3/tree/unlocked-fps-rw-dhewm3-1.5.2

Peace ✌️

Stradex:
> > > I decided to take matters into my own hands and ported Stradex's source into dhewm3 1.5.2 .
> > > To anyone landing here from pcgamingwiki or wherever, just check out my repo: https://github.com/simonedibilio/dhewm3/tree/unlocked-fps-rw-dhewm3-1.5.2
> > > Peace ✌️
> > 
> > 
> > Thanks. After comparing it's clear how Stradex's patches are much smoother than com_fixedTic -1. And you made RoE work too!
> 
> And - as of saturday - The Lost Mission too! :)

Hi! Thanks a lot for doing it. I've been really busy lately sadly and I am not finding time to work with Doom 3 stuff anymore :(

But it makes me happy to see that some of the stuff I did still serve some purpose, I am sorry for not porting it to Dhewm3-1.5.2 myself but I am glad you did.

quake1982:
> I decided to take matters into my own hands and ported Stradex's source into dhewm3 1.5.2 .
> 
> To anyone landing here from pcgamingwiki or wherever, just check out my repo: https://github.com/simonedibilio/dhewm3/tree/unlocked-fps-rw-dhewm3-1.5.2
> 
> Peace ✌️

Sir, you are doing God's work, much appreciated for this contribution. However, there are two major bugs:
1. switching mods (e.g. from based DOOM3 to ROE) causes "black screen" in a short term (from 5 to 25 seconds), which means the game menu can not be shown immediately (still the console can be shown in the screen).
2. The gravity gun in ROE or Lost Mission can not work immediately (probably in 10-30 seconds)if you first grab it or after cut-scenes.
I think those bugs must be triggered by the uncapped fps.
Cheers!

simonedibilio:
> > I decided to take matters into my own hands and ported Stradex's source into dhewm3 1.5.2 .
> > To anyone landing here from pcgamingwiki or wherever, just check out my repo: https://github.com/simonedibilio/dhewm3/tree/unlocked-fps-rw-dhewm3-1.5.2
> > Peace ✌️
> 
> Sir, you are doing God's work, much appreciated for this contribution. However, there are two major bugs:
> 
>     1. switching mods (e.g. from based DOOM3 to ROE) causes "black screen" in a short term (from 5 to 25 seconds), which means the game menu can not be shown immediately (still the console can be shown in the screen).
> 
>     2. The gravity gun in ROE or Lost Mission can not work immediately (probably in 10-30 seconds)if you first grab it or after cut-scenes.
>        I think those bugs must be triggered by the uncapped fps.
>        Cheers!

hey! thank you for taking the time to report these issues; I'd ask - if you don't mind - to move the discussion to the "issues" section on my repo, so that we can keep track of these problems there :)

Muleke:
I am really glad to see that there is still interest in making DOOM 3 run with a high refresh rate, just like in BFG Edition!

Running TDM with a high refresh rate and FPS is a dream come true, so I'm looking forward to having that implemented in dhewm3!

Terepin:
Since 1.5.3 is out, can this be ported to the newest version as well?

glKarin:
> > For people with screens that doesn't support more than 60hz, use r_swapInterval 0 (need to close and open the game again after doing it)
> 
> entering `vid_restart` in the console might suffice :)
> 
> I just remembered that there already was some discussion about >60fps at #230 and also [#15 (comment)](https://github.com/dhewm/dhewm3/issues/15#issuecomment-521099340) - revelator says he has a timing-related fix from DarkMod, so it might be worth checking out what they did (unfortunately, his modified dhewm3 doesn't seem to be on Github).
> 
> In the discussions there was something about Doom3 not properly running at 60fps, but 62.5, and it seemed that this was due to time being measured in integer(!) milliseconds.
> 
> ```c++
> const int USERCMD_HZ			= 60;			// 60 frames per second
> const int USERCMD_MSEC			= 1000 / USERCMD_HZ; // 16
> ```
> 
> A frame should be 16.666.. ms, but the code uses 16 everywhere - and 1000/16 = 62.5 (instead of the desired 60) - that (combined with vsync) might lead to micro stuttering, I guess? And millisecond resolution might cause even more problems at higher framerates (=> lower frametimes, just 4,166.. ms at 240fps).
> 
> I wonder if your changes here are affected by these problems? Might be worth checking out what darkmod and D3BFG do?
> 
> I'm not sure if that makes sense or is even possible in Doom3, but do you think we could continue to run the game logic at 60Hz, so scripts and (most of?) the code in the game DLL don't have to be changed, but render at a different rate (and probably update at least the camera so looking around feels smooth)? It _sounds_ like it should work at multiples of 60Hz (like 120Hz), but what if someone tries running the game at 90Hz or 144Hz or 165Hz - will stuttering get even worse then? If running the game code that way (async), we might be able to modify the rest of the engine to either use **micro**second for time, or maybe keep milliseconds but put them in a float or double for additional precision? I haven't investigated this topic in Doom3, so this is just ideas :)

https://github.com/glKarin/com.n0n3m4.diii4a/commit/1f62e474695e163509db780c0a3a7aa84a63736a
I am using some TheDarkMod's source about FPS in DOOM3, it seems look like well. 
Current, if enable com_fixedTic, firstperson model animation will be faster. I do not also test any more.
And I think using cvars as engine -> game instead of changing idGame::RunFrame function for keeping SDK/mod's compation.

loc199x:
still Waiting for the fps unlocked ver 1.5.3 since this is the only version that support gamepad

DanielGibson:
I created a pull request based on this with additional fixes: https://github.com/dhewm/dhewm3/pull/584

PR #260, opened 2019-10-26
Update Mac source

MaddTheSane:
Remove the rsrc file (unused)
`-[NSString cString]` is deprecated. Use `-[NSString UTF8String]` instead.

PR #302, opened 2020-07-19
Fix Cmake on macOS

dnicolson:
This pull request fixes the `cmake` step on macOS, currently this message is displayed:
```
CMake Error at CMakeLists.txt:958 (add_executable):
  Cannot find source file:

    <path>/dhewm3/neo/sys/osx/Doom 3.rsrc
```

The `Doom 3.rsrc` file was removed recently in https://github.com/dhewm/dhewm3/pull/260.


DanielGibson:
Thanks!
No idea why that file was removed in #260; as I don't own a Mac I hadn't tested it.

PR #310, opened 2020-08-27
[do not merge] Make the client async

Yamagi:
Hi,
@DanielGibson asked me if I could have a look at dhewm3s timing code and estimate if it's possible to make the client asynchronous, i.e. render as many frames as possible without breaking the game logic. This is a very first try. The approach taken is the same as in Yamagi Quake II. The client is split into two time zones, **renderframes** for the renderer and **clientframes**  for everything else.

General observations:
* The timer subsystem is very inaccurate, about +/- 3 milliseconds offset per 16 milliseconds.

What's working:
* I've played through Mars City without any obvious problems.
* Most (I haven't found a broken one) GUIs including the PDA.
* The animations seem to be okay.
* The sound is okay and in sync with the animations and videos.
* Scripts aren't obviously broken at least.

What's not working:
* The game crashed once with `Assertion 'node1->object == NULL && node2->object == NULL' failed.` in BTree.h:442. I'm unable to reproduce it. The coredump was broken, so no chance to further investigate it.

What's missing:
* Logic that slows down things a little bit currently we're burning the CPU in a tight spinloop.
* cvars to configure client- and render frames per second.
* A cvar to switch the async stuff off.
* Windows and OS X are broken.
* A lot of testing.

The big question is if this is the way to go or if it would be better to make the timer subsystem more accurate and split the global `TRIGGER_EVENT_ONE` timer (which is now unused) into two distinct times. One for the client and one for the renderer. In any case it is possible to make dhewm3 asynchronous and it looks easier than I initially thought.


Yamagi:
I just forced pushed the branch:

* Code cleanup.
* Unbreak Windows and OS X.

To make it clear. The goal of this pull request is to split the client into at least two time zones and allow the addition of more time zones at a later time. The goal is not to solve the 60 hz problem. That will be done (if ever) at a later time.

Yamagi:
I've added two cvars:
* `com_asyncClient` to switch between the synchronous and asynchronous client. Defaults to `0`, synchronous client. The code for the synchronous client is unchanged from Vanilla Doom 3, there's no emulation like in YQ2.
* `com_renderFPS` the limit the number of frames rendered when `com_asyncClient` is `1`. Defaults to `300`.
Both cvar can be set at runtime without restart. I've played through the Alpha Laps, the game seems to run fine in async mode.

DanielGibson:
just a quick untested win32 build of this branch rebased on current master: [dhewm3_1.5.1_RC2+async_win32.zip](https://github.com/dhewm/dhewm3/files/5382354/dhewm3_1.5.1_RC2%2Basync_win32.zip)

PR #333, opened 2020-11-26
Calculate EFX occlusion as in Creative's EFX-Util

bibendovsky:


DanielGibson:
Thanks for these pull requests! :-)

I hope I get around to test them this weekend.

If you're going to to make more changes like this and #332, I think it would be ok to put them all in the same branch (and thus the same pull request)?

bibendovsky:
No, I don't have plans to add another pull requests for EFX.
If it necessary I can close those two PRs and create the new one with all changes.

DanielGibson:
No, it's ok, just if you had adjusted more calculations to match creatives behavior (I don't know how many are affected) having like 5 more PRs would be a bit tedious - but these two are absolutely manageable ;-)

DanielGibson:
So with these changes, does it sound more like it used to in Vanilla Doom3 with EAX on Creative Soundcards?
(I don't really know what that sounded like, I used to play on Linux where those effects - and Doom3's OpenAL-support in general - weren't available)

bibendovsky:
Unfortunately it's unknown for me either - I don't own appropriate hardware.

DanielGibson:
Thanks for these improvements!

Issue #328, opened 2020-11-15
Stuck on Ladder in Delta Reactor Room


jyudoki:
Just encountered a game stopping bug in Delta Reactor Room (the one with lost souls). At first I had trouble walking up the initial set of stairs without jumping, with my character getting stuck. Next, when I try to go up the ladder to the next level, I get stuck and cannot proceed no matter what. 

I've looked this up and found that it's a random bug that happens with both vanilla and BFG edition. Here's a video of it from someone else:

https://www.youtube.com/watch?v=zrsW5taL88c

Could this be isolated and fixed perhaps? Attached my affected quicksave:

[QuickSave.zip](https://github.com/dhewm/dhewm3/files/5541708/QuickSave.zip)

Edit: It appears perhaps on level load all the ladders and stairs are loaded wrong so that it makes traversal difficult/impossible until the level is reloaded.

HarrievG:
I encountered this same bug too.,
Not sure it is an engine bug since you show that it even happens on the console port and original released bins.
That being said, the loading part about the ladders is interesting. Did you debug this yourself? if you can point to what is in your opinion the offending code, i can take a look.

HarrievG:
tests to perform:

- [x] Load from supplied save game. 
so far unable to reproduce from save game in 32bit win32 debug
- [ ] normal load from start of map
- [ ] DevLoad from start of map,

HarrievG:
[ladder_load_relase32.txt](https://github.com/dhewm/dhewm3/files/5832447/ladder_load_relase32.txt)
Attaching an save game load which is made from the quicksave above, and saved infront of the first ladder encountered

HarrievG:
I have done some searching, and it appears that the ladder bug is common, but only in this level. 
It also seems to be persistent from saves, but i cannot reproduce it with the supplied one. 

That ladder bug, its quite famous, and would love to get my hands on a save game that persistently reproduces this. 
For now, i dont think this should hold back the next release.

DanielGibson:
Did you also test with release binaries?
Maybe the bug doesn't happen with (unoptimized) debug binaries..
If it indeed only happens in release mode, you could try CMake's `RelWithDebInfo` release type

HarrievG:
Yes , I tried 32 bit release and debug.
Could you give a try at reproducing it? Then at least we can be sure the save game is unfortunately not the golden egg we are looking for..

HarrievG:
[...]

PR #342, opened 2021-01-26
Fix some warnings

turol:
I got tired of warnings and fixed some of the more spammy ones.

### DIFF
--- a/neo/idlib/geometry/Surface.h
+++ b/neo/idlib/geometry/Surface.h
@@ -123,7 +123,9 @@ idSurface::idSurface
 ID_INLINE idSurface::idSurface( const idDrawVert *verts, const int numVerts, const int *indexes, const int numIndexes ) {
 	assert( verts != NULL && indexes != NULL && numVerts > 0 && numIndexes > 0 );
 	this->verts.SetNum( numVerts );
-	memcpy( this->verts.Ptr(), verts, numVerts * sizeof( verts[0] ) );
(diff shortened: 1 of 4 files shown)


HarrievG:
I have to say, that although with modern compilers, removing all the memcpy's to assignments does not give an performance decrease, on older platforms it might.. 

There are some things, like replacing memcpy with loops with assignments from which again it is difficult to say what the compiler will make if it. 

Most of this comes down to how intelligent the compiler is on bringing it back to an optimised form. 

Have you compared the assembly? 

Tbh, I have the same feeling as you, about the compiler warnings, but i am trained to be very diligent with potenional performance increases as the compiler is NOT god. Being verbose is some at places intentional. (Math classes most of the time) And sometimes you have to silence it with an compile flag, instead of rewriting the code..

That all being said, I think most of your changes are oké, but **I hope that someone with _more_ experience can look at this from an performance perspective and weed out the risky ones and ease the mind of the performance mongers**..

turol:
No, I didn't inspect the assembly. I did benchmark it on a Raspberry Pi 4 just now though. No statistical difference in either load times (the save from #328) or timedemo performance on demo1.

turol:
Rebased.

1.5.1 it out now, can this be merged?

turol:
@DanielGibson ping

DanielGibson:
oh right, sorry

Issue #344, opened 2021-01-31
Fix issues with script-state in savegames and argsizes in scripts


DanielGibson:
As it turned out when debugging #328, the fix for #303 broke savegames.
Fixing this is practically impossible (without breaking #303 again, which might be the lesser of two evils in the short term).

## The original problem

To recap, the problem in #303 was that if you have a script where method **a** calls method **b** before **b**'s implementation (=> it is only known at all because of its prototype/declaration in the class/object definition), it reserves the wrong amount of space on  the stack for the passed function parameters, because their size is only calculated when parsing the implementation, not when parsing the prototype, see `idCompiler::ParseFunctionDef()`: https://github.com/dhewm/dhewm3/blob/280c95b52c5d39a4aec8ac033189d84d37f54c84/neo/game/script/Script_Compiler.cpp#L2147-L2173

## Initial fix and why it caused (other) problems

[My fix](https://github.com/dhewm/dhewm3/commit/c4c72363523fc65e635c7ad7ae86247b4f082340) was to already set the function parameter types and calculate their combined size when parsing the prototype.
Problem with this is, that it also allocations definitions for those parameters (`gameLocal.program.AllocDef( type->GetParmType( i ), type->GetParmName( i ), def, false );`), and those definitions get a number (to identify them in savegames and such) that is just incremented each time `AllocDef()` is called. As now this is done earlier and in another order than before, those definition numbers (`int idVarDef::num`) are different than they were before, which breaks  the savegames.
Specifically, `idProgram::CalculateChecksum()` returns a different value than before (and this checksum is saved in savegames), so `idProgram::Restore()` returns `false`, so `idGameLocal::InitFromSaveGame()` also returns `false` which causes `idSessionLocal::ExecuteMapChange()` to abort loading the savegame and restart the map from the beginning (though it seems like player stats like health and ammo are preserved).

## Possible improved fix and why that still has issues

Now this alone is fixable: In `idCompiler::ParseFunctionDef()` I can put the `if ( !CheckToken( "{" ) ) {` check back on top, and before returning I can calculate and set `func->parmTotal` (and probably also set `func->parmSize[i]`), *but not* do the parameter definition part (that calls `AllocDef()`) and that way the right sizes are used everywhere but the definitions are allocated in the same order as before so they don't change.
However, it turned out that the calculated checksum is still different, though there's much fewer differences in the statements.

[...]

PR #346, opened 2021-02-07
Fix script checksum in savegames, enhance savegames, fix clipping bugs in delta1

DanielGibson:
Now savegames from before c4c72363523fc65e635c7ad7ae86247b4f082340 can be loaded properly again.
Bumped the `BUILD_NUMBER` and changed the savegameformat a bit by adding information about the dhewm3 build and system the savegame was created on (of course compatibility with old savegames persists).

See #344 for more information about the issues


DanielGibson:
Also added a fix for #328

PR #350, opened 2021-02-15
add support to build non-bundled version for macos

plasticle:
I have added support for NATIVE_EXECUTABLE command line switch that affects only mac builds.
By adding -DNATIVE_EXECUTABLE when running cmake configuration, a native mac binary is built as stand-alone, and does not get put into a bundle. This allows for easier integration with potential build pipelines (like if one wants to add a Homebrew formulae, through which -core discourages the use of bundled applications, and there are currently no hosted builds of the bundled application that could be used to integrate into the -cask formulae)

### DIFF
--- a/neo/CMakeLists.txt
+++ b/neo/CMakeLists.txt
@@ -173,7 +173,7 @@ endif()
 if(MSVC)
 	# This is required for tools on windows.
 	find_package(MFC)
-	
+
 	if(TOOLS AND NOT MFC_FOUND)
 		message(SEND_ERROR "MFC ('Microsoft Foundation Classes for C++') couldn't be found, but is needed for TOOLS!")
 		message(FATAL_ERROR "If you're using VS2013, you'll also need the 'Multibyte MFC Library for Visual Studio 2013': https://www.microsoft.com/en-us/download/details.aspx?id=40770 (VS2015 and 2017 include that in the default MFC package)")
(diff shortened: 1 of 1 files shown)


neo/CMakeLists.txt
```diff
@@ -958,13 +963,13 @@ endif()
...
 
 if(CORE)
-	add_executable(${DHEWM3BINARY} WIN32 MACOSX_BUNDLE
+	add_executable(${DHEWM3BINARY} WIN32 ${bundle}
 		${src_core}
 		${src_sys_base}
 		${src_sys_core}
 		${src_editor_tools}
 	)
-
```
DanielGibson, line 967: what's the point in all these whitespace changes?

neo/CMakeLists.txt
```diff
@@ -226,11 +226,16 @@ if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID STREQUAL "Clang")
 		add_definitions(-DIOAPI_NO_64)
 	elseif(APPLE)
 		add_definitions(-DMACOS_X=1)
+		if (NATIVE_EXECUTABLE)
```
DanielGibson, line 229: whouldn't it make sense to add a proper `option()` for this?

DanielGibson: ??

PR #354, opened 2021-03-08
fix: cdkey check now properly disabled in clients

Stradex:
Using #ifdef was a problem because even if ID_ENFORCE_KEY_CLIENT was set to 0, the #ifdef will still compile anything inside that block. _(#ifdef is not the same as #if)_

I am doing this pull request because it is a problem I faced before. Every person who wants to play dhewm3 multiplayer, as client, always needs to introduce a cdkey even if this is a GPL release.


DanielGibson:
makes sense, thanks for the patch!

Issue #330, opened 2020-11-16
Chaingun sound loops when pressing escape


darkcharmander:
When shooting the Chaingun, the sound will loop whenever you open the menu while shooting. It will stop afterwards.
Attached is a small video which clearly shows this issue. (I had to zip it, because Github refuses MP4...)
[Chaingun Sound Glitch.zip](https://github.com/dhewm/dhewm3/files/5548692/Chaingun.Sound.Glitch.zip)

DanielGibson:
I think this is fixed, can you try out the testbuild linked in https://github.com/dhewm/dhewm3/pull/363 ?
Thanks!

darkcharmander:
I've tested the mentioned build, and the glitch is no longer present. Good job!

PR #363, opened 2021-04-08
Use stb_image and stb_vorbis instead of libjpeg/libogg/vorbis(file); Sound fixes

DanielGibson:
Got rid of libjpeg by replacing it with stb_image.h - this should make Linux binaries more portable, because it works around that stupid issue that Ubuntu's standard libjpeg is libjpeg-turbo in jpeg8 mode, while most other distros (incl. debian) ship libjpeg-turbo in jpeg6.2 mode.
By the way, libjpeg wasn't really used anyway, as neither Doom3, RoE or any of the mods I checked ship any JPEGs... but if any mod uses them after all, it'll continue to work, thanks to stb_image.

Also got rid of libogg and libvorbis(file) by using stb_vorbis instead - they're not nearly as annoying as libjpeg, but fewer external dependencies are still better.

While I was working on the sound code, I also fixed that issue of looping sounds continuing to play when entering the menu, see #330.
Another issue that got fixed is that `s_decompressionLimit` is regarded when loading oggs again, i.e. if the sound is shorter than 6 seconds (or whatever that CVar is set to), they're decoded on load and kept uncompressed in main memory, instead of streaming them from the ogg file on each play. Shouldn't make a noticeable difference, but might improve performance a bit on slower machines.


DanielGibson:
While all these changes seem to work for me, they could need more testing to make sure I didn't break anything.

Here's a testbuild for Windows: [dhewm3-1.5.2pre-reduce-deps.zip](https://github.com/dhewm/dhewm3/files/6275656/dhewm3-1.5.2pre-reduce-deps.zip)

Thanks in advance!

Yamagi:
Ive played through hell1.map. No obvious problems.

DanielGibson:
I added another thing: the `s_alReverbGain` CVar can be used to reduce the applied  gain when "EAX" (actually EFX) is enabled.
The default value is `1.0` (which sounds like it did before), lower values like `0.5` make the effect more subtle (higher values are not possible).
Here's a Windows build with all these changes: [dhewm3-1.5.2pre-reduce-deps-reverbgain.zip](https://github.com/dhewm/dhewm3/files/6299043/dhewm3-1.5.2pre-reduce-deps-reverbgain.zip)

DanielGibson:
Another testbuild for Windows, working around an OpenAL bug that prevented EAX/EFX per-room settings from being applied immediately (but only after going to the menu and back):
[dhewm3-1.5.2pre-reverbgain-try2.zip](https://github.com/dhewm/dhewm3/files/6307979/dhewm3-1.5.2pre-reverbgain-try2.zip)

DanielGibson:
squashed the loop sound bugfix commit into the "Use stb_vorbis instead of libogg and libvorbis(file)" commit and modified the s_alReverGain commit to default to 0.5 instead of 1.0 because I think that's closer to what most people want (1.0, or the reverb as it was before, was considered way too strong by many who subsequently disabled EAX/EFX completely).

PR #364, opened 2021-04-09
Fix some uninitialized member variables

turol:
Found with Valgrind. These might have triggered unpredictable behavior in sound or rendering.


DanielGibson:
thanks!

PR #369, opened 2021-04-16
Refactor idGameEdit

turol:
Split `idGameEdit` into base class `idGameEditBase` which is in the main binary and `idGameEdit` which is only referenced in the game libraries.

This was an attempt to fix UBSAN compilation. The main binary now compiles with UBSAN but the game library does not. It requires vtables of things which only exist in the main binary. I think the way to make it work will be to implement static linking of game library. So this might not even be necessary. Thoughts?


DanielGibson:
Sorry, somehow forgot about this.

TBH, if this doesn't really fix the UBSAN problems (as the game lib still can't be built) I don't really see the point of this change.

Implementing static linking of the game lib seems like the way to go, and shouldn't be too hard (see d3wasm) - though in case of dhewm3 it of course needs to be optional (enabled by a CMake option that's off by default).

turol:
Well there's also the FIXME comment (that I seem to have forgotten to remove) that says this should be done.

@HarrievG do you want to incorporate this into your changes?

DanielGibson:
> @HarrievG do you want to incorporate this into your changes?

Please don't - this breaks the Game API (=> makes mod DLLs incompatible), without solving any problem (that old FIXME isn't worth it)

turol:
I was under the impression that he needed to change this class anyway.

DanielGibson:
As far as I see #399 doesn't touch Game.h

HarrievG:
I'll check it out later this night or this week, will rebase, and make it
ready.

HarrievG:
_"As far as I see #339 doesn't touch Game.h"_
The debugger changes are not public, but stashed locally.
When Daniel has approved this pr, i am going to unstash my debugger changes and merge it with this one. 

_"If we break the game API in the future for some other reason we could think about incorporating this change"_
If i recall correctly; i did _not_  change the Game API, i only changed the Editor API so the debugger hooks could be exposed without adding deps

motorsep:
> When Daniel has approved the 4k fixes pr, i am going to unstash my debugger changes and merge it with this one and make a seperate pr

Well, we are waiting :D

turol:
Rebased and removed the fixed FIXME comment.

neo/framework/Game.h
```diff
@@ -239,8 +239,87 @@ class idProgram;
...
+public:
+	virtual						~idGameEditBase( void ) {}
+
+	// These are the canonical idDict to parameter parsing routines used by both the game and tools.
+	virtual void				ParseSpawnArgsToRenderLight( const idDict *args, renderLight_t *renderLight ) = 0;
+	virtual void				ParseSpawnArgsToRenderEntity( const idDict *args, renderEntity_t *renderEntity ) = 0;
+	virtual void				ParseSpawnArgsToRefSound( const idDict *args, refSound_t *refSound ) = 0;
+
+	// Animation system calls for non-game based skeletal rendering.
+	virtual idRenderModel *		ANIM_GetModelFromEntityDef( const char *classname ) = 0;
```
gl0527, line 253: I would expect `const` getters.

neo/framework/Game.h
```diff
@@ -239,8 +239,87 @@ class idProgram;
...
+	// Player methods.
+	virtual bool				PlayerIsValid() const = 0;
+	virtual void				PlayerGetOrigin( idVec3 &org ) const = 0;
+	virtual void				PlayerGetAxis( idMat3 &axis ) const = 0;
+	virtual void				PlayerGetViewAngles( idAngles &angles ) const = 0;
+	virtual void				PlayerGetEyePosition( idVec3 &org ) const = 0;
+
+	// In game map editing support.
+	virtual const idDict *		MapGetEntityDict( const char *name ) const = 0;
+	virtual void				MapSave( const char *path = NULL ) const = 0;
```
gl0527, line 311: :ambulance: :warning: Virtual functions and default parameters should not be used together.
Virtual functions bind dynamically, but default function parameter values bind statically.
So if you implement this function in a derived class with a different default parameter, and you call it through a base pointer without parameter, it will use the default parameter specified in the base class, not the one in the derived class!

neo/framework/Game.h
```diff
@@ -239,8 +239,87 @@ class idProgram;
...
+	// Entity selection.
+	virtual void				ClearEntitySelection( void ) = 0;
+	virtual int					GetSelectedEntities( idEntity *list[], int max ) = 0;
+	virtual void				AddSelectedEntity( idEntity *ent ) = 0;
+
+	// Selection methods
+	virtual void				TriggerSelected() = 0;
+
+	// Entity defs and spawning.
+	virtual const idDict *		FindEntityDefDict( const char *name, bool makeDefault = true ) const = 0;
```
gl0527, line 282: See my comment on line 311.

neo/framework/Game.h
```diff
@@ -239,8 +239,87 @@ class idProgram;
...
+	virtual void				EntitySetAxis( idEntity *ent, const idMat3 &axis ) = 0;
+	virtual void				EntityTranslate( idEntity *ent, const idVec3 &org ) = 0;
+	virtual const idDict *		EntityGetSpawnArgs( idEntity *ent ) const = 0;
+	virtual void				EntityUpdateChangeableSpawnArgs( idEntity *ent, const idDict *dict ) = 0;
+	virtual void				EntityChangeSpawnArgs( idEntity *ent, const idDict *newArgs ) = 0;
+	virtual void				EntityUpdateVisuals( idEntity *ent ) = 0;
+	virtual void				EntitySetModel( idEntity *ent, const char *val ) = 0;
+	virtual void				EntityStopSound( idEntity *ent ) = 0;
+	virtual void				EntityDelete( idEntity *ent ) = 0;
+	virtual void				EntitySetColor( idEntity *ent, const idVec3 color ) = 0;
```
gl0527, line 300: Second parameter is copied.

turol:
@gl0527 This PR was just about splitting up the class so it stops breaking UBSAN. If you have improvement suggestions make them over at #379.

PR #376, opened 2021-05-06
Resolve building built-in tools with x64 MSVC

raynorpat:
This pull request fixes build issues when building the built-in tools with 64bit MSVC builds


DanielGibson:
Wow, thanks a lot!

neo/tools/common/MaskEdit.cpp
```diff
@@ -60,7 +60,7 @@ LRESULT CALLBACK MaskEdit_WndProc ( HWND hWnd, UINT msg, WPARAM wParam, LPARAM l
 
 		case WM_DESTROY:
 			delete edit;
-			SetWindowLong ( hWnd, GWL_WNDPROC, (LONG)wndproc );
+			SetWindowLongPtr ( hWnd, GWLP_WNDPROC, (LONG)wndproc );
```
DanielGibson, line 63: Missed this when merging this: pretty sure the `(LONG)` cast should be `(LONG_PTR)` instead.
There's more similar issues and there's also compiler warnings for it.
@raynorpat Can you look at your changes again and try to fix those issues?

raynorpat:
Absolutely

PR #377, opened 2021-05-07
Additional LONG to LONG_PTR fixes for the tools

raynorpat:
This also has some minor cleanups in regards to some of the warnings.

### DIFF
--- a/neo/tools/comafx/CDIB.cpp
+++ b/neo/tools/comafx/CDIB.cpp
@@ -340,8 +340,8 @@ void CDIB::ExpandBlt(int nXDest,int nYDest,int xRatio,int yRatio,CDIB& dibSrc,in
 
 void CDIB::Expand(int nXDest,int nYDest,int xRatio,int yRatio,CDIB& dibSrc,int xSrc,int ySrc,int  nSWidth,int nSHeight)
 {
-int xNum,yNum,xErr,yErr;
-int nDWidth,nDHeight;
+	int xNum,yNum,xErr,yErr;
+	int nDWidth,nDHeight;
 
 	nDWidth = nSWidth*xRatio;
 	nDHeight = nSHeight*yRatio;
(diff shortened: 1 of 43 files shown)


DanielGibson:
Thanks, this is much better, but some fixes were still missing.
I did most of them, see https://github.com/DanielGibson/dhewm3/commits/64bit-tool-fixes

The only thing still missing is this ugly hack here: https://github.com/DanielGibson/dhewm3/blob/5dc9b69374eb44ccf739fb6e5a3a09f3a682b189/neo/tools/guied/GEWindowWrapper.cpp#L72-L94
It's not as straightforward, because it insists on storing a pointer in an idWinVar (using idWinInt), but there is no idWinVar type for pointers.. thinking about putting it into a idWinStr

DanielGibson:
I also (hopefully) fixed rvGEWindowWrapper, by converting the pointer to a hex-string and stuffing that in a idWinStr.

Here's 64bit test binaries with both raynorpats and my fixes: [dhewm3-1.5.2pre-win64-tools.zip](https://github.com/dhewm/dhewm3/files/6445233/dhewm3-1.5.2pre-win64-tools.zip)

motorsep:
- GUI Editor seems to work fine.
- Particle Editor seems to work fine except it crashed when I did Save As (I modified default particle emitter, clicked on Save As, specified particle emitter name, specified file name, clicked save/ok(?) button, got asked twice that file doesn't exist and if I want to create new one, after confirming twice Particle Editor closed silently).

motorsep:
> Do you know if it's the same in 32bits?

Not in SE2 (I just tested). Works as expected.

motorsep:
Light Editor works, along with saveLights cmd

DanielGibson:
Thanks for testing, I merged this now, the remaining issues can be fixed in later commits.

HarrievG:
@motorsep I think i have seen the particle editor crashing in 32b too. could you try the same steps in dhewm3 32b?
Also, im not sure to what extend comparing to 32b SE2 is really a good comparision.. Best thing is to just use the same program, then you are 100% sure.

PR #381, opened 2021-05-11
Functional Game Debugger

HarrievG:
- fixed game debugger

### DIFF
--- a/neo/tools/debugger/DebuggerWindow.cpp
+++ b/neo/tools/debugger/DebuggerWindow.cpp
@@ -868,7 +868,8 @@ void rvDebuggerWindow::ResizeImageList(int& widthOut, int& heightOut)
 
 	GetTextMetrics(dc, &tm);
 	int height = mZoomScaleDem ? (tm.tmHeight * (float)mZoomScaleNum / (float)mZoomScaleDem)  : 16;
-	int width = mZoomScaleDem ? (s16 * (float)mZoomScaleNum / (float)mZoomScaleDem) : 16;
+	height *= scaling_factor;
+	int width = mZoomScaleDem ? (s16 * (float)mZoomScaleNum / (float)mZoomScaleDem) : s16;
 
 	ImageList_Destroy(mTmpImageList);
 	mTmpImageList = ImageList_Create(width, height, ILC_COLOR | ILC_MASK , 0, 2);
@@ -915,7 +916,7 @@ int rvDebuggerWindow::HandleCreate ( WPARAM wparam, LPARAM lparam )
 	dc = GetDC ( mWndScript );
 	GetTextMetrics ( dc, &tm );
 	ZeroMemory ( &lf, sizeof(lf) );
-	lf.lfHeight = tm.tmHeight;
+	lf.lfHeight = tm.tmHeight * Win_GetWindowScalingFactor( mWndScript );
 	strcpy ( lf.lfFaceName, "Courier New" );
 
 	SendMessage ( mWndScript, WM_SETFONT, (WPARAM)CreateFontIndirect ( &lf ), 0 );


neo/idlib/Lib.cpp
```diff
@@ -35,7 +35,7 @@ If you have questions concerning this license or the applicable additional terms
 #include <unistd.h>
 #endif
 
-#include <SDL_endian.h>
+#include <SDL2\SDL_endian.h>
```
DanielGibson, line 38: this is double-wrong:
1. no backslashes in includes! (think of non-windows platforms)
2. `#include <SDL_endian.h>` really should work, if not something is set up wrong on your system

neo/tools/debugger/DebuggerScript.cpp
```diff
@@ -91,7 +87,7 @@ appropriate for the file being loaded.  If the script cant be compiled
 the loading of the script fails
 ================
 */
-bool rvDebuggerScript::Load ( const char* filename )
+bool rvDebuggerScript::Load ( const char* filename)
```
DanielGibson, line 90: TBH removing the space before `)` doesn't look that nice, esp. when leaving the space after `(` ;)

HarrievG: yes will fix that.
If you have seen more whitespace bugs, let me know!

neo/idlib/Lib.cpp
HarrievG: omg! that should not have gone in

DanielGibson:
Thanks for resolving the issues I mentioned!
I just realized that the dedicated server doesn't build anymore: 
> Error	LNK2019	unresolved external symbol "void __cdecl DebuggerServerCheckBreakpoint(class idInterpreter *,class idProgram *,int)" (?DebuggerServerCheckBreakpoint@@YAXPEAVidInterpreter@@PEAVidProgram@@H@Z) referenced in function "public: virtual void __cdecl idCommonLocal::DebuggerCheckBreakpoint(class idInterpreter *,class idProgram *,int)" (?DebuggerCheckBreakpoint@idCommonLocal@@UEAAXPEAVidInterpreter@@PEAVidProgram@@H@Z)	dhewm3ded

> Error	LNK2019	unresolved external symbol "void __cdecl DebuggerClientInit(char const *)" (?DebuggerClientInit@@YAXPEBD@Z) referenced in function SDL_main	dhewm3ded	C:\dev\dhewm3\build-vs2017_64\win_main.obj	1

neo/tools/debugger/DebuggerClient.cpp
```diff
@@ -187,19 +213,27 @@ Handle the DBMSG_BREAK message send from the server.  This message is handled
...
 	// Filename
-	MSG_ReadString ( msg, filename, MAX_PATH );
+	msg->ReadString ( filename, MAX_PATH );
 	mBreakFilename   = filename;
 
+#if D3_SIZEOFPTR == 4 // 32 bit
+	int ptr32b = msg->ReadInt();
+	mBreakProgram = (idProgram*)ptr32b;
+#else
+	int64_t ptr64b = msg->ReadInt64();
```
DanielGibson, line 233: I wonder if
1. this program pointer sent around is even useful and used
2. we should always send an int64 so the network protocol is the same on 32bit and 64bit

HarrievG: I have changed to always use the 64_t. that will be that.
But, i might be able to remove the usage/sharing of this pointer all together. I have to change some functions to resolve the target program from a file name, i believe that is how its done originally.

Also, this actually forces the client and server  to run on the same machine to be able to debug.. where it would be nice to be able to debug crossplatform between machines.

HarrievG:
> Thanks for resolving the issues I mentioned!
> I just realized that the dedicated server doesn't build anymore:
> 
> > Error	LNK2019	unresolved external symbol "void __cdecl DebuggerServerCheckBreakpoint(class idInterpreter *,class idProgram *,int)" (?DebuggerServerCheckBreakpoint@@YAXPEAVidInterpreter@@PEAVidProgram@@h@Z) referenced in function "public: virtual void __cdecl idCommonLocal::DebuggerCheckBreakpoint(class idInterpreter *,class idProgram *,int)" (?DebuggerCheckBreakpoint@idCommonLocal@@UEAAXPEAVidInterpreter@@PEAVidProgram@@h@Z)	dhewm3ded
> 
> > Error	LNK2019	unresolved external symbol "void __cdecl DebuggerClientInit(char const *)" (?DebuggerClientInit@@YAXPEBD@Z) referenced in function SDL_main	dhewm3ded	C:\dev\dhewm3\build-vs2017_64\win_main.obj	1

This is a problem. i have not tested ded, completly slipped my mind. 
Must fix.

HarrievG:
white space fixes needed. in d3xp

DanielGibson:
This is great progress, thanks!

> With some work it could be made that it the debugger server is always build and the client only when ID_ALLOW_TOOLS is defined

This would be needed to debug scripts on non-Windows (over network)

neo/d3xp/script/Script_Interpreter.cpp
```diff
@@ -176,135 +178,166 @@ void idInterpreter::Reset( void ) {
...
-idInterpreter::GetRegisterValue
-
-Returns a string representation of the value of the register.  This is
-used primarily for the debugger and debugging
-
-//FIXME:  This is pretty much wrong.  won't access data in most situations.
-================
-*/
-bool idInterpreter::GetRegisterValue( const char *name, idStr &out, int scopeDepth ) {
+bool idInterpreter::GetRegisterValue(const char* name, idStr& out, int scopeDepth) {
```
DanielGibson, line 181: unnecessary (but at least consistent) whitespace change; might be preferable to fix the doc comment instead of deleting it

neo/d3xp/script/Script_Interpreter.cpp
```diff
@@ -176,135 +178,166 @@ void idInterpreter::Reset( void ) {
...
 
 	out.Empty();
 
-	if ( scopeDepth == -1 ) {
+	if (scopeDepth == -1) {
 		scopeDepth = callStackDepth;
 	}
 
-	if ( scopeDepth == callStackDepth ) {
+	if (scopeDepth == callStackDepth) {
```
DanielGibson, line 197: a lot more unnecessary whitespace changes around here

for the future, could you configure VS to disable auto-formatting (or to only do it for new lines you add, not when modifying existing code)?

neo/d3xp/script/Script_Interpreter.cpp
```diff
@@ -176,135 +178,166 @@ void idInterpreter::Reset( void ) {
...
+	}
 
 	case ev_string:
-		if ( reg.stringPtr ) {
+		if (reg.stringPtr) {
 			out = "\"";
 			out += reg.stringPtr;
 			out += "\"";
-		} else {
+		}
```
DanielGibson, line 339: your autoformatter settings are real weird - either you put the braces on the same line as `else` or not, but this looks asymmetric :-P

HarrievG:
Yes going to figure out not only how to turn it off, but also see if I can find some profile that can be submitted too.

HarrievG:
this should be better. but, maybe its an idea to use clang format?

HarrievG:
- [> > With some work it could be made that it the debugger server is always build and the client only when ID_ALLOW_TOOLS is defined
> 
> This would be needed to debug scripts on non-Windows (over network) ]

HarrievG:
i think i am missing an compiler flag?

if i try to build ded on windows it complains

`D:\ID4\d3wasm\neo\sys\stub\openal_stub.cpp(143): error C2491: 'alSourceUnqueueBuffers': definition of dllimport function not allowed
D:\ID4\d3wasm\neo\sys\stub\openal_stub.cpp(145): error C2491: 'alSourcePlay': definition of dllimport function not allowed
D:\ID4\d3wasm\neo\sys\stub\openal_stub.cpp(146): error C2491: 'alSourcePause': definition of dllimport function not allowed`

HarrievG:
@DanielGibson 
With a modified al_stub, i am able to compile dedicated on windows

neo/d3xp/GameEdit.cpp
```diff
@@ -1146,3 +1146,56 @@ void idGameEdit::MapEntityTranslate( const char *name, const idVec3 &v ) const {
...
+
+
+/***********************************************************************
+
+  Debugger
+
+***********************************************************************/
+
+bool idGameEdit::IsLineCode( const char *filename, int linenumber ) const {
+	static idStr fileStr = idStr( MAX_PATH );
```
DanielGibson, line 1158: why is this `static` and why `idStr( MAX_PATH )` ?

DanielGibson:
> With a modified al_stub, i am able to compile dedicated on windows

what kind of modification did you have to do?

DanielGibson:
> D:\ID4\d3wasm\neo\sys\stub\openal_stub.cpp(143): error C2491: 'alSourceUnqueueBuffers': definition of dllimport function not allowed 

this kind of problem *should* be prevented by:
https://github.com/dhewm/dhewm3/blob/master/neo/sound/snd_local.h#L32-L37

HarrievG:
redef the macros so no import macro was there.
Actually sort what you show me.

Curious why it doesnt work.
On a side node, why not let cmake set that define?

neo/d3xp/GameEdit.cpp
HarrievG: this function does not allocate when that file string is made static, otherwise it does, and resizes also, thats the only reason i can think of now. 
bit silly optimize with an strcmp in there... removing it.

DanielGibson:
Somehow I can't get this to work. Of course the problem could be that I have no idea how to use this thing.

I start (64bit) dhewm3 with `./dhewm3.exe +set com_enableDebuggerServer 1`
Then I start a map
Then I enter `debugger` in the console and the debugger is started.
The debuggers "Output" says
```
Dhewm3 Script Debugger v1.0

Connected...
```
And that's all. The other tabs (Console, Scripts, Call Stack, ...) are empty, clicking the "Break" button doesn't do anything either.

Am I doing something wrong?

motorsep:
@DanielGibson Then you need to open one of the scripts you want to debug (whatever is running on the map, including player's stuff) and start debugging it. At least that's how original debugger works. Btw, SE2 64bit even though tools were updated to 64bit, debugger crashes when I open scripts. Works fine in 32bit.

HarrievG:
Nope you don't have to open a script. The dhewm3 dBuggger will open the
correct one itself and should crash in 64 bit as demonstrated :)

I will ping you disc today lets see what is going on.

-Harrie

motorsep:
@HarrievG Sorry, but it makes no sense - if map has 20 scripts, how does debugger know which one you want to debug and load that script ?

HarrievG:
Magic :)


 No in all seriousness; if you press pause, the runtime breaks on the first script line it executes after recieving the break command.  When the runtime breaks, it tells the debugger on which file and what line that was.

I don't lie motorsep... Im inviting you to to compile the latest 64 bit and try it out ;)


The thing Daniel is testing , is cross platform debugging.

Tbh, I think it's just mixing windows and unix paths.

DanielGibson:
> The thing Daniel is testing , is cross platform debugging.

Nope, I tested on Windows, everything running on the same machine - I'd try cross platform once I get that working

I'll ping you later, I'm still in the process of waking up ;)

DanielGibson:
- [ ] when hitting a breakpoint, mousegrab should be released (properly?), with `GLimp_GrabInput()`
- [ ] when running the game from the debugger (Debug -> Run or run button), it should pass`+set com_enableDebuggerServer 1` to the started game
- [ ] when first starting the game (with debugger enabled) and then the debugger from the game (by entering `debugger` in the console), it doesn't work (see post above)

Nice to have:
* list of all breakpoints so you can temporarily disable them without navigating to the file

HarrievG:
Ready for final rev. 👯

neo/framework/Common.h
```diff
@@ -265,6 +277,7 @@ class idCommon {
 		// it returns true if we're currently running the doom3 demo
 		// not relevant for mods, only for game/ aka base.dll/base.so/...
 		FT_IsDemo = 1,
+		FT_CheckDebuggerBreakpoint,
```
DanielGibson, line 280: please document it in a short comment, like I did for `FT_IsDemo`

neo/tools/debugger/debugger.cpp
```diff
@@ -163,9 +167,16 @@ bool DebuggerServerInit ( void )
 		gDebuggerServer = NULL;
 		return false;
 	}
+
```
DanielGibson, line 170: at the beginning of `DebuggerServerInit()`, please add an early return for `if ( gDebuggerServer != NULL )`, so we don't accidentally create rvDebuggerServer multiple times, e.g. if `com_enableDebuggerServer` thinks it has changed but hasn't (not sure if that actually happens, but better safe than sorry)

neo/tools/debugger/debugger.cpp
```diff
@@ -163,9 +167,16 @@ bool DebuggerServerInit ( void )
 		gDebuggerServer = NULL;
 		return false;
 	}
+	
+	com_enableDebuggerServer.ClearModified( );
```
DanielGibson, line 171: if we returned early because of some error or whatever, this cvar remains modified.
not sure this is desirable (would retry each frame until it succeeds, if it ever does..)

DanielGibson (changes requested):
This will need some changes and fixes (right now it won't even compile d3xp.dll, I think)

neo/d3xp/Game_local.cpp
```diff
@@ -1350,7 +1370,7 @@ void idGameLocal::InitFromNewMap( const char *mapName, idRenderWorld *renderWorl
 idGameLocal::InitFromSaveGame
 =================
 */
-bool idGameLocal::InitFromSaveGame( const char *mapName, idRenderWorld *renderWorld, idSoundWorld *soundWorld, idFile *saveGameFile ) {
+bool idGameLocal::InitFromSaveGame( const char *mapName, idRenderWorld *renderWorld, idSoundWorld *soundWorld, idFile *saveGameFile, int activeEditors) {
```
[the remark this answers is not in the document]
HarrievG: Uhm, it does! that should not be the case i thought i removed all and had compiled d3xp too. 
Looking into this

neo/d3xp/Game_local.cpp
```diff
@@ -302,6 +302,21 @@ void idGameLocal::Clear( void ) {
...
+static DebuggerArgs_t * userDebuggerArgs;
+static bool ( *checkDebuggerBreakpointFnPtr )( void ) = NULL;
+bool IsGameDebuggerActive( idInterpreter *interpreter, idProgram *program, int instructionPointer ) {
+	bool ret = false;
+
+	if ( interpreter != nullptr && program != nullptr ) 	{
+		userDebuggerArgs->interpreter = interpreter;
+		userDebuggerArgs->program = program;
+		userDebuggerArgs->instructionPointer = instructionPointer;
+		ret = checkDebuggerBreakpointFnPtr ? checkDebuggerBreakpointFnPtr( ) : false;
```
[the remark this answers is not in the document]
HarrievG: i thought this is what that user arg was for? 
Will make this arguments of the function itself.

About the naming, i dont want to another function to expose through this mechanism, so ill make naming consistent.

neo/d3xp/script/Script_Interpreter.cpp
HarrievG: yes will fix comment!

neo/d3xp/script/Script_Interpreter.cpp
HarrievG: damnit! i did figure this out man i fixed this! very anoying, i would prefer we use an cross IDE solution so that is consistent 
Like i asked last time? clang format for example?

neo/d3xp/script/Script_Interpreter.cpp
HarrievG: yea again man see provious comment. something is wierd.
this behaviour can easily be controlled by an 3rd party cross IDE solution

neo/d3xp/script/Script_Interpreter.cpp
```diff
@@ -1001,10 +1004,9 @@ bool idInterpreter::Execute( void ) {
 		// next statement
 		st = &gameLocal.program.GetStatement( instructionPointer );
 
-		if (gameLocal.editors & EDITOR_DEBUGGER) {
-			common->DebuggerCheckBreakpoint(this, &gameLocal.program, instructionPointer);
-		}
-		else if (g_debugScript.GetBool()) {
+		if ( !IsGameDebuggerActive( this, &gameLocal.program, instructionPointer )
+			&& g_debugScript.GetBool( ) )
```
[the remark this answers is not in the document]
HarrievG: ai. no., ill fix this by adding an extra function, like it was.

neo/framework/GameCallbacks_local.h
```diff
@@ -40,9 +40,12 @@ If you have questions concerning this license or the applicable additional terms
...
 
 	typedef void (*ReloadImagesCallback)(void* userArg, const idCmdArgs &args);
-	ReloadImagesCallback reloadImagesCB;
-	void*                reloadImagesUserArg;
+	ReloadImagesCallback	reloadImagesCB;
+	void*					reloadImagesUserArg;
 
+	typedef void ( *CheckBreakpointCallback )( void *userArg, idInterpreter *interpreter, idProgram *program, int instructionPointer );
+	CheckBreakpointCallback		checkBreakPointCB;
+	void*						checkBreakPointUserArg;
```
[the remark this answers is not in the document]
HarrievG: I made an attemt to make/expose two functions, but stepped back to one.

Tbh. for me, this review ends here, and ill need to double check more. It was obviously to late in the night.

neo/game/script/Script_Interpreter.cpp
```diff
@@ -274,30 +287,55 @@ bool idInterpreter::GetRegisterValue( const char *name, idStr &out, int scopeDep
...
+				return true;
+
+			case ev_float:
+				out = va( "%g", *( reinterpret_cast<float *>( &obj->data[ reg.ptrOffset ] ) ) );
+				return true;
+				
+			case ev_string:	{
+				const char* str;
+				str = reinterpret_cast<const char*>( &obj->data[ reg.ptrOffset ] );
+				if ( !str ) {
```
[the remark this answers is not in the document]
HarrievG: thb, ill flag this in code with an comment, this needs deeper investigation but is not important enough to delay the debugger intergatre.

neo/tools/debugger/debugger.cpp
HarrievG: im not sure what you mean, you did this part . 
AFAIK, there is already an check earlier that makes sure the debugger server isnt launched twice.

neo/tools/debugger/debugger.cpp
HarrievG: i think we settled on this being the desirable behaviour.

neo/framework/GameCallbacks_local.h
```diff
@@ -40,9 +40,12 @@ If you have questions concerning this license or the applicable additional terms
 struct idGameCallbacks {
 
 	typedef void (*ReloadImagesCallback)(void* userArg, const idCmdArgs &args);
-	ReloadImagesCallback reloadImagesCB;
-	void*                reloadImagesUserArg;
+	ReloadImagesCallback	reloadImagesCB;
+	void*					reloadImagesUserArg;
```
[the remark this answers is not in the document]
HarrievG: tabs no spaces?

neo/d3xp/Game_local.cpp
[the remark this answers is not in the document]
DanielGibson: user arg was mostly meant for an object that belongs to the engine and that the implementation (*in the engine*) of the exposed function uses - it should be opaque to the game dll, like "this is just another parameter I need to pass to the function, I don't know or care what it is".
in this particular case, you just wouldn't use a user arg at all, because you're only calling global functions anyway

but "normal" args that really come from the game.dll should be passed as regular function arguments

neo/d3xp/script/Script_Interpreter.cpp
DanielGibson: formatting tools are not an option - they will format *all* code and thus will fuck up `git blame` for all code.

ideally editors should only (re)format lines you've actually changed and not surrounding ones or whatever has been happening here

neo/tools/debugger/debugger.cpp
DanielGibson: are you sure?
a common error case will be that `gDebuggerServer->Initialize()` fails because port `27980` is still blocked, for example when running two dhewm3 instances.
in that case gDebuggerServer will be newly allocated and then deleted again in every frame

neo/tools/debugger/debugger.cpp
DanielGibson: there's a `if ( com_editors & EDITOR_DEBUGGER ) return false;` check, but I think that `com_editors & EDITOR_DEBUGGER` is only set if the debugger-server is actually connected to the debugger-client (at least that bit is cleared when disconnecting) - and obviously it can run without the client being connected

neo/game/script/Script_Interpreter.cpp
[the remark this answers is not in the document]
HarrievG: this is actually consistent with the ev_string below, and i dont think it means much in script to know if an value is NULL or empty

neo/d3xp/script/Script_Interpreter.cpp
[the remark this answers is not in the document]
HarrievG: Ah no, this is actually not true.
It only returns if the debugger is active, not if an breakpoint is hit. ill clarify this.

neo/framework/Game.h
```diff
@@ -310,11 +316,44 @@ class idGameEdit {
 	virtual int					MapGetEntitiesMatchingClassWithString( const char *classname, const char *match, const char *list[], const int max ) const;
 	virtual void				MapRemoveEntity( const char *name ) const;
 	virtual void				MapEntityTranslate( const char *name, const idVec3 &v ) const;
-
 };
 
 extern idGameEdit *				gameEdit;
 
+// In game script Debugging Support
+class idGameEditExt : public idGameEdit {
```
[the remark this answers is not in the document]
HarrievG: all other mods, derived from this source tree, will not have the debugger implemented the script_interpreter.cpp so it wont run. if they try to use idGameEditExt, it wont compile, they have to implement the needed functions first.

HarrievG:
> * when first starting the game (with debugger enabled) and then starting the debugger from the game (by entering `debugger` in the console), it doesn't work (see post above)
> * [ ]  if possible, don't break the Game (SDK) API

all of this is done, except point 1 and no breakpoint list.

neo/tools/debugger/DebuggerServer.cpp
```diff
@@ -197,11 +220,11 @@ bool rvDebuggerServer::ProcessMessages ( void )
 
 			case DBMSG_STEPOVER:
 				mBreakStepOver = true;
-				mBreakStepOverDepth = mBreakInterpreter->GetCallstackDepth ( );
-				mBreakStepOverFunc1 = mBreakInterpreter->GetCallstack()[mBreakInterpreter->GetCallstackDepth()].f;
-				if ( mBreakInterpreter->GetCallstackDepth() > 0 )
+				mBreakStepOverDepth = ((idGameEditExt*) gameEdit)->GetInterpreterCallStackDepth(mBreakInterpreter);
```
DanielGibson, line 223: can we run into this code (or any other code in this source file that uses `idGameEditExt`) if a (mod) game.dll that doesn't implement idGameEditExt is loaded?

neo/framework/Game.h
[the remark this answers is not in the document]
DanielGibson: What I mean is, do you ensure that all that code in DebuggerServer.cpp doesn't try to cast the `idGameEdit` to `idGameEditExt` when playing a mod that doesn't implement it?

neo/framework/Common.h
```diff
@@ -265,6 +271,15 @@ class idCommon {
 		// it returns true if we're currently running the doom3 demo
 		// not relevant for mods, only for game/ aka base.dll/base.so/...
 		FT_IsDemo = 1,
+		// the function's signature is bool fn(void) - no arguments.
+		// it returns true if the game debugger is active
+		// relevant for mods.
+		FT_DebuggerActive,
```
DanielGibson, line 277: as far as I can tell this is never used?

neo/framework/Common.cpp
```diff
@@ -3291,14 +3320,18 @@ bool idCommonLocal::GetAdditionalFunction(idCommon::FunctionType ft, idCommon::F
 			// don't set *out_userArg, this function takes no arguments
 			return true;
 
+		case idCommon::FT_UpdateDebugger:
+			*out_fnptr = (idCommon::FunctionPointer)updateDebugger;
+
```
DanielGibson, line 3325: it might be a bit hacky, but if it turns out that you need to know whether the mod DLL supports debugging (`idGameEditExt`), you could set some global `bool` here (only mod DLLs implementing `idGameEditExt` will use `FT_UpdateDebugger`) and check that at appropriate places in the debugger server

DanielGibson (changes requested):

neo/tools/debugger/DebuggerServer.cpp
HarrievG: No, no code that uses the idGameEditExt can run if CheckBreakpoints() is not called.
If an debugger clients connects, then this will blow up spectacularly .
Maybe this should be prevented, i think i will start off with an hacky global like you proposed below.

neo/framework/Common.cpp
HarrievG: yeah thx. i forgot :P

neo/tools/debugger/DebuggerServer.cpp
DanielGibson: yeah, that's what I meant in https://github.com/dhewm/dhewm3/pull/381#discussion_r661338407 ;)

DanielGibson:
This looks very good, thanks a lot!

I hope to test it in the next days and then merge it

neo/tools/debugger/DebuggerServer.cpp
HarrievG: yea i failed your point the first time, did see it the second time ;) thx for poining this out twice!

neo/tools/debugger/DebuggerServer.cpp
```diff
@@ -241,43 +272,59 @@ Send a message with no data to the debugger server.
...
 	long lineNumber;
 	long id;
-	char filename[MAX_PATH];
+	char filename[2048]; // DG: randomly chose this size
 
 	// Read the breakpoint info
-	onceOnly   = MSG_ReadBits ( msg, 1 ) ? true : false;
-	lineNumber = MSG_ReadInt ( msg );
-	id		   = MSG_ReadInt ( msg );
+	onceOnly   = msg->ReadBits ( 1 ) ? true : false;
```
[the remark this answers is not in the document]
DanielGibson: (damn, I wrote this last night but forgot to submit the review - maybe I should stick to normal comments)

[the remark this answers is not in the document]
HarrievG: will remove it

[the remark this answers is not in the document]
DanielGibson: does the debugger client not offer a "break only once" functionality that is supposed to be communicated with this bit?
in that case it should probably be implemented instead of removed? or is it implemented already just differently?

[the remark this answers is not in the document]
HarrievG: I am not sure why this is here. it is not used, and on GLP release it has never been functional. but that last bit doesnt say much since it wasnt functional all togheter. 
Anyway, maybe it was used originally to determine the next breapoint that should hit when you pressed pause? @motorsep do you have an idea?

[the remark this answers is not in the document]
DanielGibson: Maybe it was meant to be used with the "Run To Cursor" feature which doesn't seem to work?
i.e. create an only-once breakpoint at the cursor position and continue - it would break there and automatically remove the breakpoint

DanielGibson:
>     * [ ]  when first starting the game (with debugger enabled) and then starting the debugger from the game (by entering `debugger` in the console), it doesn't work (see post above)
> 

this doesn't work - the `debugger` command should automatically set `com_enableDebuggerServer 1` so the launched debugger can actually connect

HarrievG:
Heeey regression! Ok going in.

HarrievG:
> > ```
> > * [ ]  when first starting the game (with debugger enabled) and then starting the debugger from the game (by entering `debugger` in the console), it doesn't work (see post above)
> > ```
> 
> this doesn't work - the `debugger` command should automatically set `com_enableDebuggerServer 1` so the launched debugger can actually connect

do you mean first starting (with debugger disabled)?

neo/tools/debugger/DebuggerServer.cpp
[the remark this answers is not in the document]
HarrievG: Don't know , also doesn't seem important enough to do something with it.

DanielGibson:
I did some improvements

HarrievG:
Nice!!! pulling it.
Yea, the line numbers would be nice. who does it? you or me?

HarrievG:
Whoomp there it is!

neo/sys/win32/rc/Debugger.rc
```diff
@@ -149,17 +149,17 @@ END
...
 CAPTION "About Script Debugger"
 FONT 8, "MS Shell Dlg", 400, 0, 0x1
 BEGIN
     LTEXT           "Script Debugger",IDC_STATIC,35,7,81,8
-    LTEXT           "Version 1.0",IDC_STATIC,35,17,41,8
+    LTEXT           "Version 1.1",IDC_STATIC,35,17,41,8
     LTEXT           "Original version by Raven", IDC_STATIC, 35,28, 134, 8
-    LTEXT           "Dhewm3 version by Harrie van Ginneken", IDC_STATIC, 35, 38, 134, 8
-    DEFPUSHBUTTON   "OK",IDOK,165,60,50,14
+    LTEXT           "Dhewm3 version by Harrie van Ginneken \n\t\t\t\tand\n\t\t\t Daniel Gibson", IDC_STATIC, 35, 38, 134, 32
```
DanielGibson, line 161: :heart:

HarrievG:
Looks good that white, 
And about the DPI, as long as nothing falls of screen after resizing and/or zooming windows, i would say: objective achieved.

DanielGibson:
I just wonder if the font size should be automatically scaled according to the dpi scaling, so it basically has a sane default "zoom level" (where-ever the font size is set)

HarrievG:
ok there it is.
But tbh, that zoom bug is starting to annoy me now... I have no idea what is going on. 
it has something to do with the num/dem usage i think..

DanielGibson:
this is worse than before - now the script text is cut off at 100% dpi scale

DanielGibson:
I think I found a better solution, so I removed your last commit, implemented another approach (seems to work) and merged the whole thing to master so we can finally close this.

If there's still issues after all they can be fixed in new pull request, but as far as I can tell it's quite usable now :)

PR #387, opened 2021-06-15
Gamma in shader

DanielGibson:
This implements gamma correction (from r_gamma and r_brightness) in the shaders, instead of using hardware gamma.
Can be switched off with `r_gammaInShader 0`

**This needs testing!** Ideally not just with the base game but also with the supported mods.

Windows binaries for testing: [dhewm3-1.5.2pre-win32-shadergamma3.zip](https://github.com/dhewm/dhewm3/files/6666948/dhewm3-1.5.2pre-win32-shadergamma3.zip)

The git branch for compiling yourself: https://github.com/dhewm/dhewm3/tree/gamma-in-shader
Source as zip: https://github.com/dhewm/dhewm3/archive/refs/heads/gamma-in-shader.zip


DanielGibson:
If on Windows the game freezes (so far only observed with *The Lost Mission*), as a workaround you can set `win_outputEditString 0` in the console (or `+set win_outputEditString 0` as commandline argument when starting dhewm3).

I'm working on a proper fix.
(This issue has nothing to do with this brightnes/gamma feature but was introduced when I replaced libogg/vorbis with stb_vorbis and added some logmessages when decoding an ogg fails - which unfortunately happens in the async thread and logging in that thread can lead to a deadlock)

DanielGibson:
Updated build fixing the aforementioned freezes: [dhewm3-1.5.2pre-win32-shadergamma3.zip](https://github.com/dhewm/dhewm3/files/6666948/dhewm3-1.5.2pre-win32-shadergamma3.zip)

Issue #323, opened 2020-11-08
Some keys are not recognized


Sol1vaN:
Hi! and thanks for this great port.

I've tested and this work awesome.

But, are some things that need to check in the keys detection.

For example, I use the keys semicolon, Win Keys, and all many other keys are not recognized. Well, Win Key is recognized, but this minimizes the game.

In CstDoom3, all the keys can be binded, even the Win key can be binded without problem.

Here's a resume of all keys that have problems:

1) RWin key (is recognized but minimizes the game)
2) Aphostrope is not recognized
3) Semicolon is not recognized
4) \ or | (the same physic key) is not recognized
5) = or EQUAL (the same physic key) is not recognized
6) CTRL and RAlt are recognized like only ONE KEY, I mean only CTRL
7) LCTRL and RCTRL are recognized like only ONE SAME KEY, I mean, CTRL

In CstDoom3 the not recognized keys in dhewm3 are detected very fine, but I noticed are detected with different symbols (not names):

1) RWin key is recognized and NOT minimizes the game
2) Aphostrope is recognized like  '
3) Semicolon is recognized like Ñ
4) \ or | (the same physic key) is recognized like Ç
5) = or EQUAL (the same physic key) is recognized like ;
6) CTRL and RAlt are recognized like diffrerent KEYS, I mean ALT or CTRL respectively
7) LCTRL and RCTRL sadly, are recognized like only ONE SAME KEY too, I mean, CTRL

but ey! the key layer detection in CstDoom3 are mostly perfect!

So, any way to bring all this stuff to dhewm3 too?

DanielGibson:
I think I have fixes for most of these issues in #397.

> LCTRL and RCTRL are recognized like only ONE SAME KEY

Somehow I fail to see how one would use the keyboard while playing when using both Ctrl keys?
I mean, if you play on WASD or ESDF or whatever, you can only reach the left Ctrl key - and if you play on the cursor keys you can only reach the right one..

> CTRL and RAlt are recognized like only ONE KEY, I mean only CTRL

So if you press Ctrl, it is recognized as Ctrl, and if you press RAlt it is also recognized as Ctrl?
I somehow would've expected RAlt to be recognized as Alt (it is on my keyboard)..

> RWin key (is recognized but minimizes the game)

[...]

PR #397, opened 2021-07-03
Support (hopefully) all keyboard keys via scancodes

DanielGibson:
If a key is pressed whichs SDL_Keycode isn't known to Doom3 (has no
corresponding K_* constant), its SDL_Scancode is mapped to the
corresponding newly added K_SC_* scancode constant.
I think I have K_SC_* constants for all keys that differ between
keyboard layouts (which is mostly printable characters; F1-F12, Ctrl,
Shift, ... should be the same on all layouts, which means that e.g.
SDL_SCANCODE_F1 always belongs to SDLK_F1 which the old code already
maps to Doom3's K_F1).
What's extra nice (IMO) is that when Doom3 requests a *localized* name
of the key (like for showing in the bindings menu), we actually use the
name of the SDL_Keycode that *currently* belongs to the scancode, and
esp. the "Western High-ASCII characters" (ISO-8859-1) supported by Doom3
like Ä or Ñ are displayed correctly.

(I already implemented a very similar hack in Yamagi Quake II and
 reused the list of scancodes)

This should fix most of the problems reported in #323

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,0 +1,178 @@
+dhewm3 Changelog
+=================
+
+Note: Numbers starting with a "#" like #330 refer to the bugreport with that number
+      at https://github.com/dhewm/dhewm3/issues/
+
+1.5.2 (WIP)
+------------------------------------------------------------------------
+
+* Gamma and Brightness are now applied in the shaders instead of by setting hardware gamma.  
+  Can be disabled (so hardware gamma is used again) with `r_gammaInShaders 0`
+* Improvements for (Windows-only) MFC-based tools:
+    - Support 64bit (thanks *raynorpat*!)
+    - HighDPI support (thanks *HarrievG*!)
+    - PDAEditor works now
+    - Additional bugfixes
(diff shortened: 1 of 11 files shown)


DanielGibson:
I added support for Right Ctrl and Right Shift keys (=> the old Ctrl and Shift now only refer to the left buttons now) and I added the `in_grabKeyboard` CVar - if you set it to `1`, the Windows key or other global keyboard shortcuts will not take you out of the game but be handled by dhewm3.

Here's a testbuild: [dhewm3-1.5.2pre-win32-scancodes.zip](https://github.com/dhewm/dhewm3/files/6758504/dhewm3-1.5.2pre-win32-scancodes.zip)

DanielGibson:
Updated Windows build with the new console-key changes: [**dhewm3-1.5.2pre-win32-scancodes2.zip**](https://github.com/dhewm/dhewm3/files/6798727/dhewm3-1.5.2pre-win32-scancodes2.zip)

PR #412, opened 2021-08-29
Fix build for Mac OS X 10.4 & 10.5

SiliconExarch:
Since I've now been able to build openal-soft 1.19.1 for Tiger and Leopard the only things that are holding back these platforms are the absence of glStencilOpSeparate from the system OpenGL headers (they do, however, provide glStencilOpSeparateATI) and the lack of SDL2.0.4 which is required for SDL_GetGlobalMouseState(). 

I've tried to resolve these issues while making minimal changes, hopefully this is up to the standards you expect :)

### DIFF
--- a/neo/renderer/RenderSystem_init.cpp
+++ b/neo/renderer/RenderSystem_init.cpp
@@ -278,7 +278,12 @@ PFNGLPROGRAMLOCALPARAMETER4FVARBPROC	qglProgramLocalParameter4fvARB;
 PFNGLDEPTHBOUNDSEXTPROC                 qglDepthBoundsEXT;
 
 // DG: couldn't find any extension for this, it's supported in GL2.0 and newer, incl OpenGL ES2.0
+// SE: only the ATI implementation is provided on Mac OS X 10.4 and 10.5
+#if (defined(__MAC_OS_X_VERSION_MIN_REQUIRED) && __MAC_OS_X_VERSION_MIN_REQUIRED <= 1050)
+PFNGLSTENCILOPSEPARATEATIPROC qglStencilOpSeparate;
+#else
 PFNGLSTENCILOPSEPARATEPROC qglStencilOpSeparate;
+#endif
 
 /*
 =================
@@ -398,6 +403,15 @@ static void R_CheckPortableExtensions( void ) {
 	if ( glConfig.twoSidedStencilAvailable )
 		qglActiveStencilFaceEXT = (PFNGLACTIVESTENCILFACEEXTPROC)GLimp_ExtensionPointer( "glActiveStencilFaceEXT" );
 
+#if (defined(__MAC_OS_X_VERSION_MIN_REQUIRED) && __MAC_OS_X_VERSION_MIN_REQUIRED <= 1050)
+	if( glConfig.glVersion >= 2.0) {
+		common->Printf( "... got GL2.0+ glStencilOpSeparate()\n" );
+		qglStencilOpSeparate = (PFNGLSTENCILOPSEPARATEATIPROC)GLimp_ExtensionPointer( "glStencilOpSeparateATI" );
+	} else {
+		common->Printf( "... don't have GL2.0+ glStencilOpSeparate()\n" );
+		qglStencilOpSeparate = NULL;
+	}
+#else
 	if( glConfig.glVersion >= 2.0) {
 		common->Printf( "... got GL2.0+ glStencilOpSeparate()\n" );
 		qglStencilOpSeparate = (PFNGLSTENCILOPSEPARATEPROC)GLimp_ExtensionPointer( "glStencilOpSeparate" );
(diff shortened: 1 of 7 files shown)


neo/sys/glimp.cpp
```diff
@@ -249,7 +249,11 @@ bool GLimp_Init(glimpParms_t parms) {
 		// try to put the window on the display the mousecursor currently is on
 		{
 			int x, y;
+#if SDL_VERSION_ATLEAST(2, 0, 4)
 			SDL_GetGlobalMouseState(&x, &y);
+#else
```
DanielGibson, line 254: you could put that whole loop that's trying to guess  the `displayIndex` and the definition of `int x, y;` into the `#if SDL_VERSION_ATLEAST(2, 0, 4)` block, and in the else case just set `displayIndex = 0;`

neo/renderer/RenderSystem_init.cpp
```diff
@@ -398,6 +403,15 @@ static void R_CheckPortableExtensions( void ) {
 	if ( glConfig.twoSidedStencilAvailable )
 		qglActiveStencilFaceEXT = (PFNGLACTIVESTENCILFACEEXTPROC)GLimp_ExtensionPointer( "glActiveStencilFaceEXT" );
 
+#if OSX_TIGER || OSX_LEOPARD
+	if( glConfig.glVersion >= 2.0) {
```
DanielGibson, line 407: I'm not sure this works.
If the glversion is >= 2.0, `glStencolOpSeparate` must exist because it's part of OpenGL 2.0 - that ...Ati function is an extension that doesn't have to exist.

Maybe a better solution is to undo all that `PFNGLSTENCILOPSEPARATEATIPROC` code and add something like
```c
#if defined(OSX_TIGER) || defined(OSX_LEOPARD)
typedef void (APIENTRYP PFNGLSTENCILOPSEPARATEPROC) (GLenum face, GLenum sfail, GLenum dpfail, GLenum dppass);
#endif
```

neo/renderer/RenderSystem_init.cpp
```diff
@@ -278,7 +278,12 @@ PFNGLPROGRAMLOCALPARAMETER4FVARBPROC	qglProgramLocalParameter4fvARB;
 PFNGLDEPTHBOUNDSEXTPROC                 qglDepthBoundsEXT;
 
 // DG: couldn't find any extension for this, it's supported in GL2.0 and newer, incl OpenGL ES2.0
+// SE: only the ATI implementation is provided on Mac OS X 10.4 and 10.5
+#if OSX_TIGER || OSX_LEOPARD
```
DanielGibson, line 282: `#if OSX_TIGER || OSX_LEOPARD` most probably produces compiler warnings when they are not `#defined` (which they certainly won't be on non-apple platforms).

Do you really want to check their value (if OSX_TIGER is 0 or 1/non-zero, this is what `#if` does), or is checking for their existence enough (`#if defined(OSX_TIGER) || defined(OSX_LEOPARD)`)?

neo/renderer/RenderSystem_init.cpp
SiliconExarch: I'd have assumed that too but the glext.h files that ship with 10.4 and 10.5 definitely only define glStencilOpSeparateATI, presumably because the only OpenGL2 capable hardware supported by those two platforms is the ATI HD2000 series. That being said, my G5 has an X1900XT which is only OpenGL 1.5 so I'm happy to go with this solution until I'm able to acquire an iMac 7,1 for testing.

DanielGibson: If you're gonna use glStencilOpSeparateATI, `glversion >= 2.0` is no valid check for that.
You should use something like `if( R_CheckExtension( "GL_ATI_separate_stencil" ) )`

I'd suggest still adding a custom typedef for PFNGLSTENCILOPSEPARATEPROC (for OSX_TIGER and OSX_LEOPARD) and use the ATI-version as a fallback if `glversion < 2.0`

DanielGibson: if you provide your own `PFNGLSTENCILOPSEPARATEPROC` typedef a nice side-effect is that you can just use that instead of `PFNGLSTENCILOPSEPARATEATIPROC` (the function signatures are identical if I didn't miss anything), which means you won't need most of the changes you did, but only the fallback to loading `"glStencilOpSeparateATI"` in the else-case of `if( glConfig.glVersion >= 2.0)`

SiliconExarch: Sounds good, I'll have a look later this evening :)

SiliconExarch:
Thanks again for the suggestions, please take a look at the most recent commits. I've decided not to implement the ATI extension for now as I can't easily test it on OS X but I've got a newer ATI GPU in my Power Mac that works under Linux so I might revisit that at a later time (and also maybe look into why all the NPCs just T-Pose in ppc64 builds...)

DanielGibson (approved):
Looks good, I think!

Are you sure you don't have the GL_ATI_separate_stencil extension?
It might be implemented on non-ATI cards as well

PR #418, opened 2021-10-18
SDL backend, macOs warning build fixes proposal.

devnexen:


DanielGibson:
ok, thanks for the patch!

@SiliconExarch can you verify that the current master state still builds on your old OSX versions?

Issue #424, opened 2021-11-16
Filesystem calls allow r/w outside game directory


tyabus:
For example, try dir "../../../", or WriteConfig "../../../../boot/grub/grub.cfg" (If anybody is stupid enough to run dhewm as root).
I think this should be fixed. (Look at darkplaces's FS_CheckNastyPath for example)

Yamagi:
It might be a good idea to prevent dhewm3 from being run under root or setuid to 0 at all. We did the same in YQ2 a long time ago: https://github.com/yquake2/yquake2/blob/master/src/backends/unix/main.c#L114

DanielGibson:
I merged a change to prevent the game from being run as root.

I'll keep this issue open because eventually I want to look at the filesystem code and find out if preventing access out of the allowed directories is feasible

r-a-sattarov:
Wouldn't it be too stupid of me to ask to implement the cmake `ALLOW_RUN_AS_ROOT` option, which will disable this protection mechanism? 😆
Of course, this option should be disabled by default.

PR #428, opened 2021-12-06
Prevent user from running dhewm3 as root.

wof8317:
This should resolve some aspect of issue #424, at least on Linux. Any feedback is welcome.


DanielGibson:
thanks!

PR #465, opened 2022-06-16
Install Linux dist files

abouvier:
Automatically install the files from https://github.com/dhewm/dhewm3/pull/395.


DanielGibson:
@Eonfge as you added those files, do you have an opinion about these changes?

Apart from that, I never tried using `make install` or similar with dhewm3, does it work as expected for the other files (executables and game libs)?

Eonfge:
That would be great! I kind of forgot about it, so it's cool to see @abouvier pick up the slack.

My only remark would it, is that it would be extra cool if CMAKE could also update the `metainfo.xml` so it has an updated list of releases. Not sure how feasible that is though... perhaps it's easier to do it manually.

PR #471, opened 2022-07-07
Remove deprecated warnings on mac for cString use.

knev:


DanielGibson:
What is the oldest OSX version this is compatible with?
dhewm3 tries to be compatible with quite old versions (down to 10.4, I think) so it can still run on PPC Macs

neo/sys/osx/macosx_misc.mm
```diff
@@ -74,7 +74,7 @@ void Sys_DoStartProcess( const char *exeName, bool dofork = true ) {
 */
 const char* OSX_GetLocalizedString( const char* key )
 {
-	NSString *string = [ [ NSBundle mainBundle ] localizedStringForKey:[ NSString stringWithCString: key ]
+	NSString *string = [ [ NSBundle mainBundle ] localizedStringForKey:[ NSString stringWithCString: key encoding:NSUTF8StringEncoding ]
 													 value:@"No translation" table:nil];
 	return [string UTF8String];
```
DanielGibson, line 79: Hmm I wonder if it's even correct to use UTF-8 here (I know this was already the case before your change).
Because in the end it's used in `idKeyInput::KeyNumToString()` which creates key names that are supposed to be used in the menu and printed in the Doom3 console etc - and those don't support UTF-8, but only ISO-8859-1 aka latin-1, so maybe this function should use `NSISOLatin1StringEncoding` somehow?
(Is this function correct anyway? Where/how is the returned string allocated? Who frees it? Because this is called from C++ code which just takes that const char* and doesn't know anything about an underlying NSString or about freeing anything. Note: I have zero experience with objective-c)

DanielGibson: I have no idea obviously, but as that function is returning `const char*` and not NSString, how does the calling C++ code know that there's anything to free?
(I don't know how  the obj-c/obj-c++ garbage collector works, maybe it's handled automagically. Or maybe it's allocated in basically a static variable that belongs to OSX_GetLocalizedString() that is reused on next call so it doesn't leak endlessly, or soemthing else entirely)

DanielGibson:
> stringWithCString:encoding: is compatible with 10.4+.
> https://developer.apple.com/documentation/foundation/nsstring/1497310-stringwithcstring

Great!

> I actually didn't look at the code allocation, because changing out to the non-deprecated version doesn't change the allocation. 

Yeah, I know this is existing behavior, I just noticed this (and the UTF8-thing) when looking at your changes :)

> It does seem like the memory is handled by an AutoReleasePool ...

I quickly googled AutoReleasePool and it seems to use some kind of reference counting - I guess that won't help when returning a `const char*` (which presumably doesn't have a reference to the pool that could be counted) to plain-C++ code? Could we have a use-after-free here that just happened to work over the years?
Could also be I'm worrying about nothing here and it all works somehow..

> Of course, if people are mapping their international keys to input sources, then perhaps NSISOLatin1StringEncoding or NSISOLatin2StringEncoding might be more fitting. 

Yep - I think it would have to be latin1, and it should ensure that for example keyboard keys with names in that range (like `Ä` on German keyboards) are displayed correctly.
I have no idea what the behavior is like right now, would need someone who uses a Mac *and* a keyboard with such keys to test..

knev:
> I quickly googled AutoReleasePool and it seems to use some kind of reference counting - I guess that won't help when returning a `const char*` (which presumably doesn't have a reference to the pool that could be counted) to plain-C++ code? Could we have a use-after-free here that just happened to work over the years? Could also be I'm worrying about nothing here and it all works somehow..

The NSStrings get allocated and from there the cStrings are passed on. Those pointers are const, when the NSString goes out of scope, then those const pointers will be invalid (if I remember correctly). Since the `[pool release]` is at the very end of the `main`, you shouldn't have any worries.

> Yep - I think it would have to be latin1 (that's what the fonts Doom3 ships use), and it should ensure that for example keyboard keys with names in that range (like `Ä` on German keyboards) are displayed correctly. I have no idea what the behavior is like right now, would need someone who uses a Mac _and_ a keyboard with such keys to test..

It should be possible to test by just changing keyboard layout. Question is how do I determine if UTF8 or Latin is the best choice?

knev:
With UTF8 I can type Swedish in the console and bind ä to a control.

DanielGibson:
Typing in the console is independent of this, that doesn't use `KeyNumToString()`.

How is the control displayed in the bindings menu?
Should look like this:
image

(though I'm not even 100% sure if that would end up using `OSX_GetLocalizedString()` or maybe `Sys_GetLocalizedScancodeName()` which should already use iso8859-1)

knev:
Yup, that was it. I used lower case though.

DanielGibson:
Great, thanks for testing!
Doesn't really matter if it works because UTF-8 mysteriously works or because it really ends up using `Sys_GetLocalizedScancodeName()` - it seems to work, and unless someone reports an issue we can just leave it like this.

DanielGibson:
Oops, looking at it more closely, it only uses OSX_GetLocalizedString() for enter, backspace, alt, insert and printscreen - so changing the keyboard layout isn't sufficient to test it, one would have to change the system language to one that uses non-ascii-keys for those keys..
Could be that German uses "Rücktaste" for Backspace and "Einfügen" for Insert, but it could also use some other name (like "Einfg" for Insert)

Anyway, let's wait for someone to complain (if this is a bug it has existed since forever and no one did complain..)

Issue #379, opened 2021-05-09
Things to do when (if) the Game API is broken anyway


DanielGibson:
Since 1.5.0 I avoided breaking the API for Mod DLLs, so **1.** Mods maintained externally (like librecoop) are compatible with all dhewm3 versions since then, and **2.** I don't have to adjust all the mods that live in the [dhewm3-sdk repo](https://github.com/dhewm/dhewm3-sdk).

However, if it needs to be broken for some (**good**) reason, here's a list of things that could be done:
* Merge #369 which refactors the idGameEdit class
* TODO: add more to this list whenever I remember something

PR #482, opened 2022-08-06
fix `passing an object that undergoes

devnexen:
default argument promotion to 'va_start' has undefined behavior`
warning.

A bool is not an appropriate type thus proposal is just to switch
arguments, pointer is fine here.


DanielGibson:
this would break the API and ABI of the game DLL

devnexen:
oh no this one of these..not awful as is but well...

DanielGibson:
added it to this list: https://github.com/dhewm/dhewm3/issues/379

PR #500, opened 2022-12-19
Fixups for a few format-security compile-time warnings

jayaddison:
As discussed at https://github.com/blendogames/quadrilateralcowboy/pull/4#discussion_r1052342877

### DIFF
--- a/neo/game/MultiplayerGame.h
+++ b/neo/game/MultiplayerGame.h
@@ -123,7 +123,7 @@ public:
 	// updates frag counts and potentially ends the match in sudden death
 	void			PlayerDeath( idPlayer *dead, idPlayer *killer, bool telefrag );
 
-	void			AddChatLine( const char *fmt, ... ) id_attribute((format(printf,2,3)));
+	void			AddChatLine( const char *fmt, ... );
 
 	void			UpdateMainGui( void );
 	idUserInterface*StartMenu( void );
--- a/neo/cm/CollisionModel_debug.cpp
+++ b/neo/cm/CollisionModel_debug.cpp
@@ -141,7 +141,7 @@ const char *idCollisionModelManagerLocal::StringFromContents( const int contents
 			if ( length != 0 ) {
 				length += idStr::snPrintf( contentsString + length, sizeof( contentsString ) - length, "," );
 			}
-			length += idStr::snPrintf( contentsString + length, sizeof( contentsString ) - length, cm_contentsNameByIndex[i] );
+			length += idStr::snPrintf( contentsString + length, sizeof( contentsString ) - length, "%s", cm_contentsNameByIndex[i] );
 		}
 	}
 
--- a/neo/d3xp/Game_local.cpp
+++ b/neo/d3xp/Game_local.cpp
@@ -1238,7 +1238,7 @@ bool idGameLocal::NextMap( void ) {
 	int					i;
 
 	if ( !g_mapCycle.GetString()[0] ) {
-		Printf( common->GetLanguageDict()->GetString( "#str_04294" ) );
+		Printf( "%s", common->GetLanguageDict()->GetString( "#str_04294" ) );
 		return false;
 	}
 	if ( fileSystem->ReadFile( g_mapCycle.GetString(), NULL, NULL ) < 0 ) {
--- a/neo/d3xp/Game_network.cpp
+++ b/neo/d3xp/Game_network.cpp
@@ -734,7 +734,7 @@ void idGameLocal::NetworkEventWarning( const entityNetEvent_t *event, const char
 	va_end( argptr );
 	idStr::Append( buf, sizeof(buf), "\n" );
 
-	common->DWarning( buf );
+	common->DWarning( "%s", buf );
 }
 
 /*
--- a/neo/d3xp/MultiplayerGame.cpp
+++ b/neo/d3xp/MultiplayerGame.cpp
@@ -3262,7 +3262,7 @@ void idMultiplayerGame::ClientStartVote( int clientNum, const char *_voteString 
 	}
 
 	voteString = _voteString;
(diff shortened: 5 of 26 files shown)


DanielGibson:
please keep in mind that for dhewm3 the changes in neo/d3xp/ must also be applied to neo/game/

neo/game/MultiplayerGame.h
```diff
@@ -123,7 +123,7 @@ class idMultiplayerGame {
 	// updates frag counts and potentially ends the match in sudden death
 	void			PlayerDeath( idPlayer *dead, idPlayer *killer, bool telefrag );
 
-	void			AddChatLine( const char *fmt, ... ) id_attribute((format(printf,2,3)));
```
turol, line 126: Why did you remove this?

jayaddison: This is an attempt to reduce GCC compile-time warnings (`-Wformat-security`, specifically).

If it seems like it's a bad/risky idea to make this change, please let me know.

It feels to me like there _is_ some risk in the way these strings are handled and processed (depending on the source of the input text) -- but those risks are considered acceptable (for performance/functionality tradeoff reasons perhaps) by the codebase, and/or are ignored at compile-time.  That's probably me making excuses, and without a full understanding of the codebase and history, but it's an attempt to provide some background.

turol: This will not fix the underlying issue and will disable a useful warning that finds other bugs. You should not do it.

jayaddison:
Closing; please do not merge.  A similar branch/changeset may be re-opened for consideration in future.

neo/game/MultiplayerGame.h
jayaddison: Thank you; I've closed this, and added no-merge notices to two related pull requests that were tracking the same set of modifications (https://github.com/iodoom/iod3/pull/3 , https://github.com/blendogames/quadrilateralcowboy/pull/4).

I'd like to understand how to address the underyling issue, and then this changeset can be revisited.

DanielGibson: It totally makes sense to disable that warning here.
The problem is that almost every instance of calling AddChatLine() uses it with a variable as format string (a localized string, e.g. `common->GetLanguageDict()->GetString( "#str_04299" )`), so
1. unlike with a string literal as format string, the compiler can't verify the arguments anyway
2. we get shittons of -Wformat-security warnings that we can't do anything about

See also [this comment](https://github.com/blendogames/quadrilateralcowboy/pull/4#issuecomment-1359550576) (and the surrounding comments)

jayaddison: (and the `AddChatLine` function definition does, itself, [apply `printf` logic](https://github.com/dhewm/dhewm3/blob/a7e4eb86449799379f1e62b8a9890abab8fa457f/neo/d3xp/MultiplayerGame.cpp#L2620-L2622) -- so the removal of the annotation doesn't remove functionality)

But: if there's an upstream/higher-level issue, even if it might take ages to resolve, I'd like to understand what that is and at least consider fixing it before moving ahead with something potentially-risky.

(from a naive re-reading about format printing bugs and problems, I feel like it's something fairly fundamental to `printf`, so... unless that has changed much in recent C++ versions, this could be a pain.  filtering out `%n` and `%x` or other sketchy patterns feels like it could be an improvement.  and/or using `vsnprintf`?  the fact that it's difficult to even determine what to do, or where the problems could be, as a newcomer I think is a kind of programming language meta-analysis insight)

turol: GCC has attribute `format_arg` for just this case. Chocolate Doom has example of it and also code for validating that the replaced format string doesn't do bad things.
https://github.com/chocolate-doom/chocolate-doom/blob/d34a98b434ccccf5bedbd799ec003551f34707d8/src/deh_str.h#L27
https://github.com/chocolate-doom/chocolate-doom/blob/d34a98b434ccccf5bedbd799ec003551f34707d8/src/deh_str.c#L369

jayaddison: Good to know, although: could we confirm what the bug/undesirable behaviours are before considering potential solutions?

(rationale: without being fairly precise about the problem scenarios, it's not possible to test/confirm that they are resolved by a change)

turol: The replacement format string must have the same output specifiers and in the same order as the original. Also they must match the actual parameters. If not, undefined behavior will result, including possible crashes and security issues.

DanielGibson: I think that's overkill.

The problem scenario is what? A .pk4 (from a mod?) overwriting strings/english.lang with format strings that don't match?

I'd be happy with just fixing the lower hanging fruit for now (like all the other compiler warnings that this PR fixed before the source branch was removed)

turol: I think mod developers/translators would appreciate if a mistake causes an error message instead of crash. But actual players are unlikely to benefit.

DanielGibson: > I think mod developers/translators would appreciate if a mistake causes an error message instead of crash

Maybe, but I think they also appreciate fewer differences between the Doom3 SDK and the dhewm3 SDK, so merging code between them is easier

DanielGibson: nevermind, I just tried this myself and having the printf-attribute on AddChatLine() only causes a compiler warning when called with a single argument (like `AddChatLine( common->GetLanguageDict()->GetString( "#str_04278" ) );`) - **not** when calling it with multiple arguments like `AddChatLine( common->GetLanguageDict()->GetString( "#str_04293" ), gameLocal.userInfo[ parm1 ].GetString( "ui_name" ) );`.

So the attribute can stay

DanielGibson:
I integrated most of your changes from this PR (and some of my own related changes) in this (for now experimental) branch: https://github.com/DanielGibson/dhewm3/tree/fix-compiler-warnings - thanks a lot!

Note that I undid removing the AddChatLine() attribute and the pragmas in File.cpp, because they weren't needed after fixing the calls with only one argument

Issue #334, opened 2020-11-28
ROE will not load


MacDoomer:
Hi all,

Currently running dhewm3 1.5.0.1304 on macOS Catalina. The base game runs fine. When I try to start ROE, it does not load with the following error:

`ERROR: Function 'map_erebus1::main' not found in script for 'call' key on worldspawn`

I made sure that all the proper files are in the d3xp folder. I've tried using all of the contents of the d3xp folder as usual, as well as only the pak000.pk4 and pak001.pk4 files. Same error each time.

Other steps taken to resolve:

- Checked the md5 hashes for the pk4 files against the hashed on the Dhewm3 main page. They match.
- Checked the location of the pk4 files. Tried both the binary directory and the ~./doom3/d3xp folder and they both produce the same result. A review of the console indicates that the Dhewm3 binary is finding the pk4s and loading them properly.

LETIshNick:
Hi,

I've just come here to report the same issue. It happens on High Sierra. [Here is the log](https://github.com/LETIshNick/errors_copypastas/commit/e85edcaa4856750bd13ec059a28590944a4e0052), hope it will be useful.

DanielGibson:
I think these are different issues, as you don't have the `ERROR: Function 'map_erebus1::main' not found in script for 'call' key on worldspawn` error.

Where did you get your dhewm3 executable? Did you compile yourself? Might be worth trying the latest git code, there have been lots of changes since 1.5.0

@MacDoomer Do you have a `d3xp.dylib` next to your `base.dylib`? Also, can you post the whole terminal output?

@LETIshNick That looks like a bug in Apples OpenGL implementation, as you can see in the backtrace it's several calls deep in system OpenGL libs when the crash happens. Maybe upgrading to a newer macOS version would help, or maybe it would make it even worse, no idea.. though I heard that dhewm3 works even on Big Sur with ARM hardware (with this little change I merged a few weeks ago: https://github.com/dhewm/dhewm3/pull/322).

LETIshNick:
@DanielGibson I've complied mine from dhewm3-1.5.1_RC2 repo. Unfortunately, the latest macOS available for my laptop is 10.13. I know about the dosdude's patcher but unfortunately this is my main machine now and I wouldn't like to tinker with it.

DanielGibson:
Ok, that sucks :-/

[...]

Issue #200, opened 2018-08-25
Fails to load first level


acowle3:
I keep getting this error: 

`ERROR: Function 'map_marscity1::main' not found in script for 'call' key on worldspawn`

What is going on with this?

acowle3:
I got it from Steam.

DanielGibson:
ok.
are there any other files in base/ than the ones listed in the aforementioned link?

acowle3:
Not originally.  But then I copied the other files found in maps, materials, and script. there are also two other files called "default.cfg" and "consolehistory.dat"

acowle3:
There's also another directory called home/.config/dhewm3 with some config files, should I take a look at them?

DanielGibson:
so there were only pak000.pk4 to pak008.pk4, not zpak000.pk4 or similar?

what operating system are you on? what version of dhewm3 are you using (and how did you install it)?

acowle3:
Only Pak000.pk4 to pak008.pk4.

I'm using Ubuntu 18.04, and I compiled it.

DanielGibson:
from latest git?

acowle3:
Yes.  Latest git.

DanielGibson:
where did you put the game data? are you running the game from the build directory?

acowle3:
I put the game data in Home/.local/share/dhewm3/base

DanielGibson:
ok. haven't tried that in a while, though it *should* work theoretically.
can you try putting it somewhere else and start the game with `./dhewm3 +set fs_basepath /path/to/gamedata/` ?
(where gamedata/ contains base/ which contains the .pk4 files)

acowle3:
When I compiled, I cannot find where it's builded, so I run it from my shell.

acowle3:
The menu boots up, just not the game itself.

DanielGibson:
it's built in the directory you built it in?
like usually you'd do something like
```
cd dhewm3
mkdir build
cd build
cmake ..
make -j8
```
then wait until it's done compiling
and then the dhewm3 executable is right there, in the build/ directory, and you can execute
`./dhewm3 +set fs_basepath /path/to/gamedata/`

please try that.

acowle3:
It still doesn't work.

DanielGibson:
Weird.
Do you know what compiler (version) was used to build it?

And can you post the whole console output from starting the game up to the error message you posted? (maybe put it into a pastebin, or a .txt file that you attach here)
[...]

Issue #90, opened 2013-10-18
Can't start Resurection of Evil


Gps2010:
After compiling and playing through Doom3, I wanted to play the expansion pack again.

Linux (openSUSE 12.3 64 bit )

I had some problems and then I copied all pks files, to the d3xp folder. ( all 13)

Then I started doom3, and it asked for my ROE cd key.
II typed it in, and all seemed fine.
Under mods I selected d3xp.
Then I click new game, and picked marine.

But during loading I get an error.

==== Processing events ====
SpawnPlayer: 0
loaded collision model models/weapons/shell1/mshell_lo.lwo
loaded collision model models/weapons/shell1/sshell_bigger.lwo
TODO: Sys_SetClipboardData
----- Game Map Shutdown -----

---

ERROR: Unknown weapon in decl 'weapon_bloodstone_passive'

---

]exit

I am totally lost on what is wrong.

If it matters I bought this game through steam, and I have played through doom3 and roe on windows 7,
now trying to play it again on openSUSE.

mkrupczak3:
On Ubuntu (not compiling from source), I found that ROE wouldn't launch. I'd get the following error:

`ERROR: Function 'map_erebus1::main' not found in script for 'call' key on worldspawn`

To fix this, I had to install the following package to obtain the missing `d3xp.so` binary:

`sudo apt install dhewm3-d3xp`

After that, everything worked!

This was in addition to the `dhewm3` and `dhewm3-doom3` packages

PR #517, opened 2023-05-10
Honour fs_game_base when defaulting to base game library.

Ionwerks:
I'll confess I haven't considered this exhaustively from all angles but having struck;

 _"main not found in script for 'call' key on worldspawn"_

...when trying to load a content-only mod on top of RoE. I wondered why it was always defaulting to the game library for "base" rather than looking for one nominated by 'fs_game_base'.

Proposed here is a correction.

Similar reports;
https://github.com/dhewm/dhewm3/issues/90
https://github.com/dhewm/dhewm3/issues/200
https://github.com/dhewm/dhewm3/issues/334

### DIFF
--- a/neo/framework/Common.cpp
+++ b/neo/framework/Common.cpp
@@ -2639,6 +2639,7 @@ idCommonLocal::LoadGameDLL
 void idCommonLocal::LoadGameDLL( void ) {
 #ifdef __DOOM_DLL__
 	const char		*fs_game;
+	const char		*fs_base;
 	char			dll[MAX_OSPATH];
 	idStr			s;
 
@@ -2650,6 +2651,10 @@ void idCommonLocal::LoadGameDLL( void ) {
 	if (!fs_game || !fs_game[0])
 		fs_game = BASE_GAMEDIR;
 
+	fs_base = cvarSystem->GetCVarString("fs_game_base");
+	if (!fs_base || !fs_base[0])
+		fs_base = BASE_GAMEDIR;
+
 	gameDLL = 0;
 
 	sys->DLL_GetFileName(fs_game, dll, sizeof(dll));
@@ -2659,7 +2664,7 @@ void idCommonLocal::LoadGameDLL( void ) {
 	if (!gameDLL) {
 		common->Printf( "\n" );
 		common->Warning( "couldn't load mod-specific %s, defaulting to base game's library!\n", dll );
-		sys->DLL_GetFileName(BASE_GAMEDIR, dll, sizeof(dll));
+		sys->DLL_GetFileName(fs_base, dll, sizeof(dll));
 		LoadGameDLLbyName(dll, s);
 	}
 


DanielGibson:
Thanks!

I looked at it briefly and it seems to make sense, will have to look more closely at it before merging though.

> Similar reports ...

Those are from people who tried to run d3xp without having a d3xp .dll/.so/.dylib, your changes won't fix the problem they had (but that's ok, their problem is only fixable by installing the d3xp lib).

neo/framework/Common.cpp
```diff
@@ -2659,7 +2664,7 @@ void idCommonLocal::LoadGameDLL( void ) {
 	if (!gameDLL) {
 		common->Printf( "\n" );
 		common->Warning( "couldn't load mod-specific %s, defaulting to base game's library!\n", dll );
-		sys->DLL_GetFileName(BASE_GAMEDIR, dll, sizeof(dll));
+		sys->DLL_GetFileName(fs_base, dll, sizeof(dll));
 		LoadGameDLLbyName(dll, s);
 	}
```
DanielGibson, line 2669: Maybe there should be an additional fallback: if the fs_game_base DLL (e.g. d3xp.dll) can't be loaded, fall back to the BASE_GAMEDIR DLL (base.dll)?

In that specific case that most likely won't work, but imagine a content-only addon (`foo`) based on the base game, and another content-only addon (`bar`) based on the former? Then you'd use `dhewm3 +set fs_game_base foo +set fs_game bar` but it should still use `base.dll` (because neither foo.dll nor bar.dll exists)

Ionwerks: Sure if you like. It crossed my mind but I decided it's not really a mechanism for stacking mods which, if they're found to work together, could be achieved by grouping the pak files in a single folder (IMHO better, in as much as it requires a conscious choice, implies compatibility, and the method is not limited to just 2).

DanielGibson: No idea if what I described above is actually done, but IMO it would be a much cleaner and safer way to make one mod base on another than copying the pak files together: This way several mods could base on `foo`, and you could have them installed all at once without having to worry about them breaking each other - if a mod says it needs `foo` as base mod, that means that it's compatible with it (ignoring that foo might get an update that breaks this..), and the mod could ship a .bat file to start that sets `fs_base_game foo` (like Lost Mission does for d3xp, though of course in that case you'll need d3le.dll anyway).

Furthermore, using a base mod instead of copying the paks into the same folder ensures the order in which the pk4 files are loaded (or traversed when looking for a file), making sure that when playing `bar` it first looks in its own pk4s, then in the `foo` pk4s, then in the base `pk4s` (well, it should do that at least, I hope this actually works, VFS implementations in games tend to be error-prone :-p).

Anyway, *if* there *are* any mods that are doing this right now, they *would* work without your patch (because if there's no mod-specific DLL it *always* falls back to `base.dll`), but **wouldn't** work with your patch, because then it'd suddenly demand `foo.dll` and not fall back to `base.dll`.

Ionwerks: It's up to you. I just think it's a very narrow case (2 content mods requiring base) which breaks down with numerous other scenarios (3+ mods, 2 content mods requiring d3xp). To my mind it's better to stick with a single methodology for mod combining (collect in one folder). Load order can be controlled by pak naming (again, a conscious choice which implies pre-tested compatibility).

DanielGibson: I'd like to have that fallback. It's just a few extra lines and won't break anything

DanielGibson:
implemented in https://github.com/dhewm/dhewm3/commit/373f59c4841fec726c672cb6aa2373d15ea2cc43

PR #518, opened 2023-05-22
Added new screenshot formats.

eezstreet:
Added the ability to change the screenshot format with a new CVar: r_screenshotFormat. This has four options: 0 (TGA), 1 (BMP), 2 (PNG), and 3 (JPG). JPG quality can be selected with a new CVar: r_screenshotJpgQuality. TGA is the default to preserve original behavior from DOOM 3.

A new dependency on stb_image_write has been added. It has been included. No additional configuration with CMake is required.

### DIFF
--- a/neo/renderer/RenderSystem.h
+++ b/neo/renderer/RenderSystem.h
@@ -229,7 +229,7 @@ public:
 	// This will perform swapbuffers, so it is NOT an approppriate way to
 	// generate image files that happen during gameplay, as for savegame
 	// markers.  Use WriteRender() instead.
-	virtual void			TakeScreenshot( int width, int height, const char *fileName, int samples, struct renderView_s *ref ) = 0;
+	virtual void			TakeScreenshot( int width, int height, const char *fileName, int samples, struct renderView_s *ref, int overrideFormat = -1 ) = 0;
 
 	// the render output can be cropped down to a subset of the real screen, as
 	// for save-game reviews and split-screen multiplayer.  Users of the renderer
--- a/neo/renderer/RenderSystem.h
+++ b/neo/renderer/RenderSystem.h
@@ -229,7 +229,7 @@ public:
 	// This will perform swapbuffers, so it is NOT an approppriate way to
 	// generate image files that happen during gameplay, as for savegame
 	// markers.  Use WriteRender() instead.
-	virtual void			TakeScreenshot( int width, int height, const char *fileName, int samples, struct renderView_s *ref, int overrideFormat = -1 ) = 0;
+	virtual void			TakeScreenshot( int width, int height, const char *fileName, int samples, struct renderView_s *ref ) = 0;
 
 	// the render output can be cropped down to a subset of the real screen, as
 	// for save-game reviews and split-screen multiplayer.  Users of the renderer
--- a/neo/renderer/RenderSystem_init.cpp
+++ b/neo/renderer/RenderSystem_init.cpp
@@ -1279,7 +1279,7 @@ Downsample is the number of steps to mipmap the image before saving it
 If ref == NULL, session->updateScreen will be used
 ==================
 */
-void idRenderSystemLocal::TakeScreenshot( int width, int height, const char *fileName, int blends, renderView_t *ref ) {
+void idRenderSystemLocal::TakeScreenshot( int width, int height, const char *fileName, int blends, renderView_t *ref, int overrideFormat ) {
 	byte		*buffer, *flippedBuffer;
 	int			i, j, k;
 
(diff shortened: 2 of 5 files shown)


DanielGibson:
Just remembered, there are other places that call `TakeScreenshot()`, with filenames that are *not* from `R_ScreenshotFilename()`:
* Session.cpp `idSessionLocal::Frame()` (for `aviCaptureMode`, passing "demos/$aviDemoShortName/$aviDemoShortName_$aviTickStart.tga" as filename)
* RenderSystem_init.cpp `R_EnvShot_f()`, passing "bla_px.tga", "bla_nx.tga" and similar as filename - and here I'd argue that it should continue using TGA, because that's the format best suited and supported for textures.

eezstreet:
I've implemented the other changes. Just a quick question on this, though:

> Using zlib for PNG deflate compression (compresses much better than stb_image_write's builtin deflate), see https://github.com/nothings/stb/commit/be21113512b4c1ab8a9fce23d5ad9ca4184bc25e
(we already use zlib for zip support, so you only need to include its header and call its functions, no messing with the buildsystem is needed)

Can you give me an example of how to do this?

DanielGibson:
> Can you give me an example of how to do this?

The commit I linked has an example in the commit message, I think that should work as it is?

eezstreet:
Thank you. I have merged in that change. We should be ready to go now.

neo/renderer/RenderSystem_init.cpp
```diff
@@ -1261,67 +1299,86 @@ Downsample is the number of steps to mipmap the image before saving it
...
-	buffer[12] = width & 255;
-	buffer[13] = width >> 8;
-	buffer[14] = height & 255;
-	buffer[15] = height >> 8;
-	buffer[16] = 24;	// pixel size
+	// The buffer is upside down, we need to flip it the right way.
+	for (i = 0; i < width; i++) {
+		for (j = 0; j < height; j++) {
+			for (k = 0; k < 3; k++) {
+				flippedBuffer[(i + j * width) * 3 + k] = buffer[(i + (height - 1 - j) * width) * 3 + k];
```
DanielGibson, line 1343: I don't think you need a flipped buffer and all this code.
just create a small buffer that can hold one line of pixels, and then use memcpy() to swap the buffer line-wise inplace.
somehow like this (untested): 
```c++
int lineSize = width*3;
byte* swapBuf = _alloca(lineSize);

for(int i=0; i < height/2; ++i) {
	byte* line1 = &buffer[i*lineSize];
	byte* line2 = &buffer[(height-i-1)*lineSize];
	memcpy(swapBuf, line1, lineSize);
	memcpy(line1, line2, lineSize);
	memcpy(line2, swapBuf, lineSize);
}
```

neo/renderer/RenderSystem.h
```diff
@@ -229,7 +229,7 @@ class idRenderSystem {
 	// This will perform swapbuffers, so it is NOT an approppriate way to
 	// generate image files that happen during gameplay, as for savegame
 	// markers.  Use WriteRender() instead.
-	virtual void			TakeScreenshot( int width, int height, const char *fileName, int samples, struct renderView_s *ref ) = 0;
+	virtual void			TakeScreenshot( int width, int height, const char *fileName, int samples, struct renderView_s *ref, int overrideFormat = -1 ) = 0;
```
DanielGibson, line 232: RenderSystem.h is part of the dhewm3-sdk, so changing a method signature here breaks the ABI - I wonder if we can avoid that somehow

neo/renderer/RenderSystem_init.cpp
eezstreet: This works. Adding! 🎉

neo/renderer/RenderSystem.h
eezstreet: I resolved this by adding a global variable, `g_screenshotFormat`, that is `extern`'d in places where it's needed.

eezstreet:
Ready for re-review and/or merging when ready.

turol:
`extern` declaration(s) should be in a header which is also included in the file with the definition. This way the compiler will warn about possible mismatches.

You should add this to #379

neo/renderer/RenderSystem_init.cpp
```diff
@@ -47,6 +47,29 @@ If you have questions concerning this license or the applicable additional terms
 #include "sys/win32/win_local.h"
 #endif
 
+#include <zlib.h>
+
+unsigned char* compress_for_stbiw(unsigned char* data, int data_len, int* out_len, int quality)
```
DanielGibson, line 52: could be static

DanielGibson:
> extern declaration(s) should be in a header which is also included in the file with the definition. This way the compiler will warn about possible mismatches

I think it should work like it is - the global variable "lives" in RenderSystem_init.cpp and is referenced as an external variable in Session.cpp (not much point in putting it into a header if it's only used in one more file, and probably shouldn't be too widely used anyway)

> You should add this to https://github.com/dhewm/dhewm3/issues/379

good point

neo/renderer/RenderSystem_init.cpp
eezstreet: Done.

DanielGibson:
Merged, thanks a lot! :)

DanielGibson:
By the way: How would you like to be credited in the changelog? as "eezstreet" or "Nick" or your full name from the commit log?

eezstreet:
“eezstreet (Nick Whitlock)” is fine. Thanks!

Sent from my iPhone

On Jun 25, 2023, at 9:42 PM, Daniel Gibson ***@***.***> wrote:

﻿

By the way: How would you like to be credited in the changelog? as "eezstreet" or "Nick" or your full name from the commit log?

PR #532, opened 2023-09-16
Added 21:9 and 32:9 resolutions

fgsfds:


neo/renderer/RenderSystem_init.cpp
```diff
@@ -536,14 +536,26 @@ vidmode_t r_vidModes[] = {
...
 	{ "Mode 15: 1920x1080",		1920,	1080 },
-	{ "Mode 16: 1920x1200",		1920,	1200 },
-	{ "Mode 17: 2048x1152",		2048,	1152 },
-	{ "Mode 18: 2560x1600",		2560,	1600 },
-	{ "Mode 19: 3200x2400",		3200,	2400 },
-	{ "Mode 20: 3840x2160",		3840,   2160 },
-	{ "Mode 21: 4096x2304",		4096,   2304 },
-	{ "Mode 22: 2880x1800",		2880,   1800 },
-	{ "Mode 23: 2560x1440",		2560,   1440 },
+	{ "Mode 16: 2048x1152",		2048,	1152 },
```
DanielGibson, line 539: These lines correspond to `r_mode N` in the config, so we can't reorder them (would break backwards compat in the configs) => new modes must be added at the end.
For displaying modes in the menu, dhewm3 has a hack that orders all modes by width (and, as secondary criteria, height)

fgsfds: Fixed

DanielGibson:
thanks!

Issue #536, opened 2023-11-03
Request:  Is Controller Support Possible?


Big-the-Fat:
I can't play with mouse and keyboard for more than a few minutes these days.  I can only properly play games with a controller.  I would love to play Dhewm3 but I have to use Doom 3 BFG.  

I would consider this an accessibility issue. :)

DanielGibson:
Yeah, it has been on my Todo list for a long time.. I hope to get around to it eventually ;)

Big-the-Fat:
Thank you. :)

RinMaru:
really would be an amazing edition allowing me to get rid of BFG one and for all.

ViperAcidZX:
I'm curious: could rumble support be implemented too if this gets controller support one day?

DanielGibson:
I'm currently working on gamepad support in this branch: https://github.com/dhewm/dhewm3/tree/controller-input

It already mostly works, but still has some issues, esp. in multiplayer.

I don't have any plans to work on rumble support, and I'm not sure how to implement it anyway (where/how to get the information when to rumble), but if someone figures it out: patches/pull requests are welcome ;)

DanielGibson:
I think it's not pretty much ready. I'll upload Windows-Binaries for testing later, if you build yourself anyway you can use the current state of the aforementioned branch.

DanielGibson:
Here's a build for Windows: [dhewm3-1.5.3pre-gamepad.zip](https://github.com/dhewm/dhewm3/files/13968587/dhewm3-1.5.3pre-gamepad.zip)

wof8317:
Just tested on my PS3 controller. I loved how the game displayed the proper name of the buttons like Cross, Triangle, etc. Stick navigation in the menus was smooth and easy. Unless you need to polish the controller code more, I'd say that it is ready to ship.

DanielGibson:
Great, thanks a lot for testing!

> I loved how the game displayed the proper name of the buttons like Cross, Triangle

Yeah, today I found out that SDL2 (from 2.0.12 on) supports detecting the kind of gamepad, so implementing this was easy enough.

I guess I'll merge it in a few days, I hope I'll get some more feedback in the meantime :)

DanielGibson:
Note: I did some small improvements for the displayed button names, especially with Nintendo-like controllers, which aren't in the windows build linked above.

Please test the build anyway, as this is a pretty cosmetic change.

DanielGibson:
[...]

PR #548, opened 2024-01-15
Modified some code to make it compilable.

wof8317:
Replaced K_JOY1 and K_JOY2 (which no longer exists in KeyInput.cpp and KeyInput.h) with its current equivalent K_JOY_BTN_SOUTH and K_JOY_BTN_EAST respectively. I do not know whether or not those are the actual names you want to replace K_JOY1 and K_JOY2, but had to make a very good guess to make the code buildable.

### DIFF
--- a/neo/ui/Window.cpp
+++ b/neo/ui/Window.cpp
@@ -736,7 +736,7 @@ const char *idWindow::HandleEvent(const sysEvent_t *event, bool *updateVisuals) 
 				*updateVisuals = true;
 			}
 
-			if (event->evValue == K_MOUSE1 || event->evValue == K_JOY2) {
+			if (event->evValue == K_MOUSE1 || event->evValue == K_JOY_BTN_EAST) {
 
 				if (!event->evValue2 && GetCaptureChild()) {
 					GetCaptureChild()->LoseCapture();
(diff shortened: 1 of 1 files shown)


DanielGibson:
oh right, forgot to commit that - thanks!

DanielGibson:
How do you like the gamepad code in general? Does it feel usable (once one has set up the bindings)?
I'm not used to playing with gamepads so it's hard to tell for me..

wof8317:
The gamepad code, the key names in particular, look very informative and easy to read, unlike the generic Joy1, Joy2, etc. Once bindings are set up and in-game, it feels very responsive, with not much input latency (I tested with an Xbox controller in Bluetooth). Of course, menu and PDA navigation is not there yet (but I'm sure that you will find a way to implement that), and there is no way to bind the Start button to open the menu right now. However, when using a mouse and hovering over a selection, pressing the A button actually selects it, which was a pleasant surprise. I feel confident that the controller code will continue to flesh out well enough to merge to the main branch.

DanielGibson:
I pushed some further changes with improvements.
Main change is that you can now move the mouse cursor with the sticks (though not in the PDA yet, because the PDA is implemented as a horrible hack - it's basically a weapon in gamecode, instead of a proper fullscreen menu :grimacing:)

DanielGibson:
now the PDA also works, but it all doesn't work that well in multiplayer mode yet.
it's turning kinda ugly :-/

wof8317:
Nice. I'll check it out and see how it feels.

wof8317:
Can't get it to compile. Error is the follwing: "'sessLocal'  wasn't declared in this scope" in the member function "virtual const char* idUserInterfaceLocal::Activate(bool, int)" in UserInterface.cpp, line 587.

wof8317:
Found the solution: include "framework/Session_local.h" to UserInterface.cpp

DanielGibson:
I just pushed another commit that already fixes this :)

DanielGibson:
I think my branch is now ready, if you want to test :)
Let's move further discussion to #536

wof8317:
Okay. I'll test it once I am home and have access to a controller. Continuing discussion at #536.

PR #549, opened 2024-01-17
Added missing include

wof8317:
UserInterface.cpp was missing the Session_local.h file that has the "sessLocal" function. Without it, compilation would have been unsuccessful.


DanielGibson:
already fixed it (sessLocal isn't needed in UserInterface.cpp anymore), thanks :)

wof8317:
No problem.

PR #553, opened 2024-02-01
Workaround for MCST-LCC compiler < 1.28 version

r-a-sattarov:
After 6f35ce045dd3ab57cac053301b1ebac4d5270501 commit compilation on mcst-lcc became problematic (compiler does not support __builtin_alloca_with_align). Added workaround for < 1.28 version.
Ref: http://mcst.ru/lcc

### DIFF
--- a/neo/sys/platform.h
+++ b/neo/sys/platform.h
@@ -185,12 +185,14 @@ If you have questions concerning this license or the applicable additional terms
 // Unix
 #ifdef __unix__
 
-#ifdef	__GNUC__
-  #define _alloca16( x )			( ({assert((x)<ID_MAX_ALLOCA_SIZE);}), __builtin_alloca_with_align( (x), 16*8 ) )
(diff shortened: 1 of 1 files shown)


ursoft (approved):

DanielGibson:
thanks!

DanielGibson:
I just found out that `__builtin_alloca_with_align()` is generally broken (memory might be released at end of block instead of end of function), so I removed it for all platforms.

Can you confirm that the regular `__builtin_alloca()` is supported by all versions of MSCT-LCC, or was that also only added in 1.28?

r-a-sattarov:
> Can you confirm that the regular `__builtin_alloca()` is supported by all versions of MSCT-LCC, or was that also only added in 1.28?

Yep, based on the [output of CE](https://ce.mentality.rip/z/bxh8Kz), this should work in any version.

PR #569, opened 2024-03-29
Fix desktop files

mnhauke:
"desktop-file-validate" reports the following error:

org.dhewm3.Dhewm3.desktop: error: value "dhewm3 "$@"" for key "Exec" in group "Desktop Entry" contains a non-escaped character '$' in a quote, but it should be escaped with two backslashes ("\\$")

see
https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#exec-variables


DanielGibson:
@Eonfge any thoughts on this?

Eonfge:
LGTM. The Flatpak didn't experience any problems with the unescaped $, but it's good to follow the official specification

DanielGibson:
thanks!

Issue #575, opened 2024-05-12
Echo of the of SSG shot became much quieter.


AntonieS:
OpenAL Soft 1.23.1, Win7 x64 SP1. Dhewm 3 1.5.3.1305.
Subj compare to 1.5.2.1503. Now it sounds unnatural.

DanielGibson:
How does it sound compared to 1.4.1?
In 1.5.0 I reduced the volume of all sounds, which (due to some sounds being configured much louder than the can be played, and the max volume being clamped before setting it in OpenAL) changed the relative volume of some sounds, see https://github.com/dhewm/dhewm3/issues/326#issuecomment-1366833004 for an explanation.

In 1.5.3 that is fixed (I clamp the volume before reducing it), so it *should* sound like in 1.4.1 and older (incl. orig Doom3 with OpenAL backend) again, except for the bug that was fixed by reducing the sounds in the first place (https://github.com/dhewm/dhewm3/issues/179), which is still fixed.

AntonieS:
1.4.1 is very bad.
1.5.2 sounds much better. But its about echo. 
Overall 1.5.3 lesser SSG echo quality but better in other: some ambient sounds, for example announcements that the passage is closed in Admin map, appears only in 1.5.3.

tyuah8:
In my opinion, the code written to reduce the volume of the sounds here:

[4567f26](https://github.com/dhewm/dhewm3/commit/4567f265392b5ab95a335f8896ff003a743990e4)

is both unnecessary and incorrect, as all it does is move the problem to a different volume, since it multiples the volume by 0.333f.

I recommend **completely restoring the original game code (not just removing the 0.333f)**, and instead, use the following in alsoft.ini:

## output-limiter:
#  Applies a gain limiter on the final mixed output. This reduces the volume
#  when the output samples would otherwise clamp, avoiding excessive clipping
#  noise.
**output-limiter** = **true**

By default, **output-limiter** is set to **false**, which causes the situation where too many loud sounds playing causes some of them to go quiet.  Setting it to **true** completely eliminates this issue.

DanielGibson:
I don't have influence on users alsoft.ini/alsoft.conf

tyuah8:
Very well, but perhaps you can make your changes optional so users have the option of doing what I just mentioned (using **output-limiter** = **true** instead of the hard-coded volume reduction, which seems to be causing other issues).

It is what I opted for with my version of the DOOM 3 engine.

[...]

PR #576, opened 2024-06-04
Integrate Dear ImGui and implement a new settings menu with it + more

DanielGibson:
This integrates the [Dear ImGui](https://github.com/ocornut/imgui) GUI toolkit and using that provides a new settings menu (which can be opened with `F10` unless that key is bound already, or by entering `dhewm3Settings` in the console).
Dear ImGui support can be disabled when building dhewm3, and is automatically disabled for SDL1.2, as it requires SDL2.

The settings menu has many options that Doom3s options menu doesn't have, like:
* Gamepad configuration and other advanced input settings
* Enhanced keybinding menu that allows an arbitrary amount of keys bound per command, instead of limiting to three
* Allow changing resolution without having to restart dhewm3 ("Apply" button does `vid_restart`)
* Additional video settings
* Additional audio settings, incl. selection of OpenAL device and enabling/disabling HRTF
* More game options like toggle crouch or show FPS
* Can be navigated also with gamepad or keyboard, instead of just with the mouse

While I implemented this I also added some features to dhewm3:
* HighDPI support (by separating mouse coordinate space from physical GL framebuffer size; needs SDL2)
* Invert Mouse look (both horizontally and vertically) with `m_invertLook`
* CVar to allow always run in single player (still drains stamina though!): `in_allowAlwaysRunInSP`
* VSync can be enabled/disabled on the fly, without restarting the renderer (still with `r_swapInterval` or in the menu, of course; needs SDL2)
* Allow enabling/disabling [HRTF](https://en.wikipedia.org/w/index.php?title=Head-related_transfer_function) with `s_alHRTF`
* `s_alOutputLimiter`: Configure OpenAL's output-limiter which temporarily reduces the overall volume when too many too loud sounds play at once, to avoid issues like clipping
* `s_scaleDownAndClamp`: Clamp and reduce volume of all sounds to prevent clipping or temporary downscaling by OpenAL's output limiter
* Dropped the dependency on zlib, now [miniz](https://github.com/richgel999/miniz) is used instead (built as part of dhewm3). This should also works around newer CMake versions fucking up their zlib library support (https://github.com/dhewm/dhewm3/issues/575#issuecomment-2145901938)

As these are a lot of changes, this could need some testing!
Here's a current build for Windows: [dhewm3-1.5.4pre-imgui3_win32.zip](https://github.com/user-attachments/files/15546318/dhewm3-1.5.4pre-imgui3_win32.zip)


DanielGibson:
Some screenshots:
d3-bindingsmenu
d3-videooptions
d3-hidpi

DanielGibson:
Thanks for testing and reporting that bug - I just pushed a commit that should fix it :)

Glad you like the new menu!

DanielGibson:
Here is an updated build for Windows: [dhewm3-1.5.4pre-imgui4_win32.zip](https://github.com/user-attachments/files/15747687/dhewm3-1.5.4pre-imgui4_win32.zip)

Since the last one there are a few fixes and one new feature:  
**The dhewm3 window is now freely resizable!** In windowed mode, you can drag the edges to resize, you can maximize (and restore) it, whatever.
Please test!

@j4reporting: Those mingw issues should be fixed, can you test the lastest code?

DanielGibson:
Damn, how did I not notice this..

Thanks for reporting (and testing!), it should be fixed now :)

DanielGibson:
try again, turns out I only tested with r_fullscreenDesktop 1 :see_no_evil:

DanielGibson:
My plan was to keep it like this.
When using the menu or Alt-Enter, `vid_restart partial` is already used, the remaining use cases are people explicitly running vid_restart in the console, and
1. they should know what they're doing (I'll probably document the vid_restart behavior in Configuration.md)
2. a full vid restart might help with weird bugs/glitches that the partial one doesn't, due to reloading everything and recreating the window
3. I don't want to break the old behavior (also because of 2.)

DanielGibson:
I reworked the resolution/windowed/MSAA section of the Video Options.

Now the CVars are set immediately and "Apply" only does `vid_restart partial`, while "Reset" resets the CVars to the values that were set when opening the menu.
This also works the other way around: Changing a CVar (in the console or other menu or r_fullscreen with Alt-Enter) is immediately reflected in the menu

I think this is more consistent with how the  rest of the menu works.

I also made "fullscreen desktop" a separate option, so one can configure windowed mode but still configure fullscreen desktop to specify what happens on Alt-Enter.

PR #578, opened 2024-06-17
Implement Soft Particles (and capturing depth buffer), based on old The Dark Mod code

DanielGibson:
Now (view-aligned) particles don't "cut" into geometry anymore, but fade instead

Can be disabled with `r_useSoftParticles 0`


DanielGibson:
Potential problem with this implementation: DeclParticle.h is part of the SDK, so adding `softeningRadius` to `idParticleStage` breaks the game ABI

RenderSystem.h is also in the SDK, but glConfig_t *really* shouldn't be used by game DLLs (they can't really access it anyway, the global `glConfig` is in tr_local.h, which is *not* part of the SDK.

DanielGibson:
I tested them in Mars City Underground, here:
r_useSoftParticles 1:
shot00077
r_useSoftParticles 0:
shot00076

and here:
r_useSoftParticles 1:
shot00074
r_useSoftParticles 0:
shot00073

(note that those "stripes" from particles "cutting" into the geometry are gone with r_useSoftParticles 1)

Sol1vaN:
this will be available on the 1.5.3 version or upcoming release?
i tryed to enable the r_useSoftParticles on 1.5.3 version and is not available
(sorry for the dumb question, but i don't understando those "branch", "pull request", "merge"

so i imagine that features only can be tested for those people that can "compile" the code pointing to these experimental "branch".

i'm not sure about what i'm saying, but i thinks is some near to the right.
i hope some day this implementation (and the 144hz can be present on the 1.5.3 version)
cheers.

DanielGibson:
1.5.3 was released three months ago, so it doesn't contain the the feature of this pull request from two weeks ago.

A pull request still being open (not merged) means that this code is not part of the official dhewm3 code yet.
And even if it's closed as merged, it's only part of releases that were created *after* the merge.

Sol1vaN:
ok i understand, i need to wait for that feature merge to the official branch.
thanks for the explanation. 😎👍

DanielGibson:
there are test binaries for windows at https://github.com/dhewm/dhewm3/pull/585#issuecomment-2204800096

DanielGibson:
I modified the code to avoid breaking the game API/ABI (by disabling the "softRadius" particle stage keyword in particle definitions, which was introduced by TDM and thus isn't used by any gamedata relevant for dhewm3 anyway), so I can merge this now.

I plan to release dhewm3 1.5.4 for Doom3's 20th birthday on August 3rd, and release a release candidate today or tomorrow or so.

The big features in that release will be the soft particles and the ImGui-based menu.

High-FPS support will *not* make it into the release, as there's no chance that it'll be done and sufficiently tested in 9 days (currently it still has some physics bugs in SinglePlayer at high framerates, and Multiplayer is still completely broken with > 60fps).

DanielGibson:
@Sol1vaN: there's a new dhewm3 release candidate (including soft particles), that could use some testing: https://github.com/dhewm/dhewm3/releases/tag/1.5.4_RC1

PR #582, opened 2024-06-21
Material addition for portal skies

BielBdeLuna:
Added a easier material name for portal skies, keeps the old also


BielBdeLuna:
related to: https://github.com/dhewm/dhewm3-sdk/pull/35

BielBdeLuna:
at least this doesn't conflict with anything, it's a far more easier recognizable name for portal skies, and you still keep the old material name too,

DanielGibson:
merged, thanks

Issue #250, opened 2019-09-12
Concept of dhewm3 running at 144Hz refresh


dezo2:
I made a custom win32 build for my testing purposes aimed to 144Hz displays, including RoE and dentonmod. After some testing I decided to share it here in case anyone of you guys wants to try it. It is by no means official proper engine fix, just an experiment if it can run smoothly at such 
refresh. I also tried to fix some issues found on the way - see bellow. There can be more of them (mainly physics), but hopefully nothing game breaking.

[dhewm3_144Hz.zip](https://github.com/dhewm/dhewm3/files/3607566/dhewm3_144Hz.zip)

To install this just copy the unzipped content over the stable 1.5.0 dhewm3 release, or merge the autoexec lines if you already use your own. Of course your display refresh has to be 144Hz before executing the game and in-game VSync must be active. The frame time for 144Hz is very close to 7ms, so that is the value I used in all binaries for the integer USERCMD_MSEC constant. With that and the com_fixedTic "1" plus VSync, the game speed looks alright. As for the CPU load, do NOT use r_finish "1" for input lag reduction, it tanks the CPU in some areas and may cause crashes in high refresh. Just leave it at "0", input lag will be lowered anyway with the shorter frame time.

Changes made in the binaries:
- all: physics workarounds to fix some machinery and moved/pushed items
- all: too fast air decreasing when outside
- RoE: replaced hard-coded 16/4 speeds in artifact slowmo ramps
- dentonmod: new weapon not selected (when switch to new is active)
- dentonmod: hud drawed on top of white or red screen overlays
- dentonmod: copied the widescreen crosshair fix from the base game
And this was adjusted in the scripts:
- all: too fast fire rate of the chaingun
- dentonmod: lost souls sometimes keep screaming after death

dezo2:
During testing I went through the whole base game and RoE expansion pack using i7-8700, GTX-1080 (drivers 431.60), 2560x1440 @144Hz with default dhewm3 Ultra setting. I didn't use the in-game MSAA, instead used cheap nVidia FXAA, which looks OK in this old game and is almost resources free. GPU was resting the whole time at about 10%..30% load and I didn't encountered any engine crashes. Surprisingly there also weren't any slowdowns due to com_fixedTic "1", not even once. So at least on this CPU the high refresh seems to be viable.

dezo2:
[...]

PR #584, opened 2024-06-26
WIP: Support >60Hz with com_gameHz

DanielGibson:
Based on @Stradex' ["Add com_gameHz to play with higher HZ/FPS" Pull Request](https://github.com/dhewm/dhewm3/pull/297), but also integrating fixes from @dezo2's [patch](https://github.com/DanielGibson/dhewm3/commit/46e519f24556fd87c87d2133896ce75c6399355d) ([see also](https://github.com/dhewm/dhewm3/issues/250)) and my own fixes.

Very much WIP, and no guarantees this will work out

----
Potential bugs that have been observed with various attempts of running dhewm3 with >60Hz and that are worth testing:

- [ ] [Crashes with crane in Alpha Labs Sector 3](https://github.com/dhewm/dhewm3/issues/230#issuecomment-502461387) (should be fixed by patching GAME_FPS and GAME_FRAMETIME in the scripts)
- [ ] Chaingun shooting too fast (should be fixed by patching CHAINGUN_FIRE_SKIPFRAMES in script)
- [ ] [Flickering particles](https://github.com/dhewm/dhewm3/issues/261) (happened with `com_fixedTic -1`, might not be relevant here?)
- [ ] [Oxygen drainage too slow or fast](https://github.com/simonedibilio/dhewm3/issues/2) (should be fixed by modifying the "air" code in idPlayer)
- [ ] [Switching Mods/game modes can cause black menu for multiple seconds to minutes](https://github.com/simonedibilio/dhewm3/issues/2)
    - I can reproduce this bug.
- [ ] [RoE gravity gun only works after 2 minutes or so](https://github.com/simonedibilio/dhewm3/issues/2) (haven't tried this yet, but this is very relevant here as this also happened with @simonedibilio's branch that's also based on Stradex' code)
- [ ] [I found if I tweak the fps higher than the monitor refresh rate, NPCs and mobs in-game will no longer interact with stairs. I think this is not a bug related to this fork, cause I found it in BFG Edition too.](https://github.com/simonedibilio/dhewm3/issues/2) (again from same bugreport on simonedibilio's branch - this sounds extremely weird)
- [ ] "There is also a lipsync issue, the animation goes a little slower than it should be." (again from same bugreport as the previous ones)
- [ ] Savegames don't work, "Script checksum didn't match" (same bugreport; hopefully fixed by 9cb3fae45da, though this will happen with all old savegames, but ones created with the latest code should keep working in the future)

### DIFF
--- a/neo/d3xp/Camera.cpp
+++ b/neo/d3xp/Camera.cpp
@@ -514,7 +514,7 @@ void idCameraAnim::Think( void ) {
 			return;
 		}
 
-		if ( frameRate == USERCMD_HZ ) {
+		if ( frameRate == gameLocal.gameFps ) {
 			frameTime	= gameLocal.time - starttime;
 			frame		= frameTime / gameLocal.msec;
 		} else {
@@ -568,7 +568,7 @@ void idCameraAnim::GetViewParms( renderView_t *view ) {
 	SetTimeState ts( timeGroup );
 #endif
 
-	if ( frameRate == USERCMD_HZ ) {
+	if ( frameRate == gameLocal.gameFps ) {
 		frameTime	= gameLocal.time - starttime;
 		frame		= frameTime / gameLocal.msec;
 		lerp		= 0.0f;
--- a/neo/d3xp/Game_local.cpp
+++ b/neo/d3xp/Game_local.cpp
@@ -1328,8 +1328,6 @@ void idGameLocal::MapPopulate( void ) {
 	// before the physics are run so entities can bind correctly
 	Printf( "==== Processing events ====\n" );
 	idEvent::ServiceEvents();
-
-	SetScriptFPS(static_cast<const float>(this->gameFps));
 }
 
 /*
@@ -1391,8 +1389,6 @@ bool idGameLocal::InitFromSaveGame( const char *mapName, idRenderWorld *renderWo
 
 	gameRenderWorld = renderWorld;
 	gameSoundWorld = soundWorld;
-
-	SetScriptFPS(static_cast<const float>(this->gameFps));
 
 	idRestoreGame savegame( saveGameFile );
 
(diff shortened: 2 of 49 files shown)


tyuah8:
There is also a bug related to the crouch rate of the player; this crouch rate uses an exponential smoothing logic and doesn't scale with higher FPS, so the player will crouch faster at higher FPS.

The original game code in Player.cpp looks like this:

```
	if ( EyeHeight() != newEyeOffset ) {
		if ( spectating ) {
			SetEyeHeight( newEyeOffset );
		} else {
			// smooth out duck height changes
			SetEyeHeight( EyeHeight() * pm_crouchrate.GetFloat() + newEyeOffset * ( 1.0f - pm_crouchrate.GetFloat() ) );
		}
	}
```

The fixed version, which properly considers FPS change for exponential smoothing, is the following:

```
	if ( EyeHeight() != newEyeOffset ) {
		if ( spectating ) {
			SetEyeHeight( newEyeOffset );
		} else {
			// https://en.wikipedia.org/wiki/Exponential_smoothing#Time_constant
			const float tau = -16.f / idMath::Log( pm_crouchrate.GetFloat() );
			const float a = 1 - idMath::Exp( -gameLocal.msec / tau );
			// smooth out duck height changes
			SetEyeHeight( EyeHeight() * ( 1 - a ) + newEyeOffset * a );
		}
	}
```
The new code I attached is what I am currently using for my local source code.  I don't have Dhewm3's source code locally, so I am unable to contribute.

tyuah8:
The same can be said about the roll and pitch rate of flying NPCs.  The fixed version, in AI.cpp `AdjustFlyingAngles()`, is the following:

```
	// https://en.wikipedia.org/wiki/Exponential_smoothing#Time_constant
	static const float tau = -16.f / idMath::Log( 0.95f );
	       const float a = 1 - idMath::Exp( -gameLocal.msec / tau );
	fly_roll  = fly_roll  * ( 1 - a ) + roll  * a;
	fly_pitch = fly_pitch * ( 1 - a ) + pitch * a;
```

DanielGibson:
Thank you! :)

DanielGibson:
oh ffs, why can't you make a PR use a different branch (or rename it) :-/

DanielGibson:
When investigating the "black screen after changing mod" issue I realized that I don't understand all of Stradex' changes, and even would like to avoid some of them (like breaking the network protocol).

I'll create a new branch (and PR for it) that is closer to @dezo2's work (but still has a CVar like com_gameHz and will modify the scripts on load so I don't have to ship new scripts) + the additional fixes for obvious bugs, but tries to be less invasive than Stradex branch.
I'll port Stradex' features as/if needed.

DanielGibson:
@tyuah8 By the way, in d3xp scaling with `gameLocal.msec` might not be exactly what you want (esp. for the player movement case), because that value gets modified for slow motion mode.

tyuah8:
I forgot to mention, and you could be right, but since I don't personally own _Doom 3: Resurrection of Evil_, all of my changes so far have been for the base game.

DanielGibson:
Ok, trying this again at https://github.com/dhewm/dhewm3/pull/585 with a slightly different approach..

PR #621, opened 2024-10-09
Support SDL3

DanielGibson:
Port dhewm3 to SDL3, while still supporting SDL2 and even SDL1.2.
Seems to work well on Linux (and probably BSD and similar), but other platforms are still TODO.
Will probably only make Windows work and try to make Mac work; no idea if SDL3 (or SDL2, for that matter..) even supports AROS.

### DIFF
--- a/neo/CMakeLists.txt
+++ b/neo/CMakeLists.txt
@@ -1224,6 +1224,9 @@ include_directories(${CMAKE_BINARY_DIR})
 include_directories(${CMAKE_SOURCE_DIR})
 
 add_library(idlib STATIC ${src_idlib})
+if(SDL3)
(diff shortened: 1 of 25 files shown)


DanielGibson:
Ok, as far as I can tell this works on Linux and Windows, and at least doesn't break the SDL2 Mac build (I don't have a Mac, so I can't properly test anything, and even testing if it builds with SDL3 by using the Github Testbuilds is currently too much work because homebrew doesn't have SDL3 yet).
Could certainly need more testing though!

DanielGibson:
Here's a Windows build for testing current dhewm3 with SDL3: [dhewm3-1.5.5pre-win32_SDL3.zip](https://github.com/user-attachments/files/17318954/dhewm3-1.5.5pre-win32_SDL3.zip)

PR #631, opened 2024-11-17
Revelator darkmod AVX* changes

revelator:
changes for adding DarkMods AVX* based culling and cpu detection as well as the changes nessesary to use them in the renderer.


revelator:
A good deal of formatting differences make these a bit larger than they have to be but i cannot do much about that.
should be no conflicts at least.

neo/idlib/math/Simd_AVX.cpp
```diff
@@ -0,0 +1,137 @@
...
+/*
+============
+idSIMD_SSE3::GetName
+============
+*/
+const char *idSIMD_AVX::GetName( void ) const {
+	return "MMX & SSE & SSE2 & SSE3 & AVX";
+}
+
+#elif defined(_MSC_VER) && defined(_M_IX86)
```
DanielGibson, line 49: why is this code (and the AVX2 code) only enabled when using MSVC on 32bit(!) x86?

DanielGibson:
Did you do any benchmarks to see how much this improves performance?

revelator:
ah damn forgot to add the enums for the new cpu types (slaps head) :/

just forgot to add them for the other targets in the middle of correcting that :) 

i did do a simple benchmark with vsync off in sikkmod with all the gruesome applied i got 130 fps in 4k.

revelator:
i dont currently have a linux box on hand to test for mistakes but i guess the buildbot in git does have one.
ill see if there are more things to look out for.

DanielGibson:
> i did do a simple benchmark with vsync off in sikkmod with all the gruesome applied i got 130 fps in 4k.

and how much without these changes?

neo/idlib/math/Simd.cpp
```diff
@@ -332,7 +327,9 @@ void TestAdd( void ) {
 	}
 	idLib::common->Printf( "====================================\n" );
 
+	// ======================
```
DanielGibson, line 330: what's the purpose of all these changes?

neo/idlib/math/Simd_AVX.cpp
```diff
@@ -35,19 +35,6 @@ If you have questions concerning this license or the applicable additional terms
 //
 //===============================================================
 
-#if defined(__GNUC__) && defined(__SSE3__)
```
DanielGibson: removing all this probably won't work either, esp. on non-x86 CPUs

revelator:
hmm compiles here on x64 atleast on msvc but try it on a linux build and let me know if it fails.

forgot to add the type for ALIGNTYPE16 its basically the same as ALIGN16 with a small difference 

```
#define ALIGN16( x )				__declspec( align( 16 ) ) x
#define ALIGNTYPE16				__declspec( align( 16 ) )
```
for msvc

```
#define ALIGN16( x )				x __attribute__ ( ( aligned ( 16 ) ) )
#define ALIGNTYPE16				__attribute__ ( ( aligned ( 16 ) ) )
```

for gnu and probably other unixes.

without the avx code changes sikkmod could barely reach above 60 fps with the heaviest settings (was playable but it did show)
my card is an older RTX 2080 ti

revelator:
happy i did not also add the GLSL renderer yet im a bit rusty in handling pull requests and coding in general, been retired for some time. Lets hope the last change clears up the build errors. :)

revelator:
looking at it now i see i could just have written it as 	
`
ALIGN16( unsigned char	FXSaveArea[512] );`

neo/sys/platform.h
```diff
@@ -38,11 +38,9 @@ If you have questions concerning this license or the applicable additional terms
...
 // NOTE: By default Win32 uses a 1MB stack. Doom3 1.3.1 uses 4MB (probably set after compiling with EDITBIN /STACK
 // dhewm3 now uses a 8MB stack, set with a linker flag in CMakeLists.txt (/STACK:8388608 for MSVC, -Wl,--stack,8388608 for mingw)
-// Linux has a 8MB stack by default, and so does macOS, at least for the main thread.
-// Anyway, a 1MB limit for _alloca() should be safe even when using it multiple times
-// in the same function or callstack.
-// If there's a risk of bigger stack allocations, Mem_MallocA() should be used instead.
-#define ID_MAX_ALLOCA_SIZE 1048576 // 1MB
+// Linux has a 8MB stack by default, and so does macOS, at least for the main thread
+// anyway, a 2MB limit alloca should be safe even when using it multiple times in the same function
+#define ID_MAX_ALLOCA_SIZE 2097152 // 2MB
```
DanielGibson, line 43: is there a specific reason you increased this to 2MB?

DanielGibson:
I recommand using the `git gui` tool to make commits.
It allows you to commit only parts of your changes, by hunk or by line(s), so you can leave out changes that aren't relevant (empty lines, lines with `// =======` etc)

nbohr1more:
From a TDM perspective, I think the bulk of the AVX changes happened in 2.06 and 2.07 but it's hard to discern how much performance impact they had since we also went "Multi-Core" ( SMP ) with game code vs the renderer in 2.06. I can say that in 2.06 the 32-bit executable was at least 25% slower than the 64-bit because we hadn't yet ported SIMD optimizations to 32-bit. 

Though I think 32-bit lost about 10% performance by virtue of going to a Dhewm3 codebase with legacy ( Pentium 4 ) SIMD features removed. ( I know that AMD 64 players instantly pounced on 2.06 as a long awaited vindication that their hardware was always better than the Pentium 4.) 

So I guess if this is successfully back-ported, we should expect a 15 to 20% performance depending on the scene. ( This presumes that other engine optimizations like our new light culling system (etc), SMP, vertex buffer changes, aren't also included in here somewhere. ). 

Maybe Rev will open a new PR to submit our asset loading optimizations? ( SIMD vectorized mipmap generation was one of the major wins in TDM 2.10 for loading time improvements. )

DanielGibson:
more optimizations from TDM are definitely welcome, preferably with one PR per feature :)

revelator:
inline problem with linux and mac intrinsics i see argh...

definatly need some more work for those but atleast the windows builds seem to go through now.
ill revert the ALIGNTYPE16 change as it is unessesary with ALIGN16.

DanielGibson:
By the way, you can also test building with GCC on Windows by using MinGW32-w64.
The Yamagi Quake II MinGW buildenv worked for dhewm3 last time I tried it, see https://github.com/yquake2/yquake2/blob/4e41dbf186d6d125b35ea338ec7472313071f0d5/doc/020_installation.md#prerequisites-on-windows-when-using-mingw
and the MinGW-specific parts of https://github.com/dhewm/dhewm3?tab=readme-ov-file#using-the-provided-windows-binaries
(you must run `cmake` in the MinGW shell so it can find its gcc executable etc)

revelator:
i do have msys2 with the mingw compilers on another pc.
will move current workdir there also to test with gcc / clang.

revelator:
calling it a night for now, its getting quite late here in denmark :)

revelator:
to late actually i have to go to the dentist in an hour xD.

reverting some macros for the SSE simd instructions so that it checks for the __SSE__ flag and adding a few new ones for the AVX* ones checking for the __AVX__ and __AVX2__  flag on gcc / clang since they are the ones making trouble atm.

for avx / avx2 the addition of -mavx -mavx2 might also be needed for gcc / clang i reckon unless checking for those flags are enough. ill try a test build with gcc when i get back :)

neo/sys/platform.h
revelator: i dont remember changing that ? maybe something from TDM carried over :S

revelator: no just checked this flag does not exist there and neither anywhere else...
no idea tbh.

DanielGibson:
I'd prefer to have the GLSL code in a separate PR

neo/renderer/tr_main.cpp
```diff
@@ -773,43 +755,51 @@ void R_TransformClipToDevice( const idPlane &clip, const viewDef_t *view, idVec3
...
-				a [ i * 4 + 0 ] * b [ 0 * 4 + j ]
-				+ a [ i * 4 + 1 ] * b [ 1 * 4 + j ]
-				+ a [ i * 4 + 2 ] * b [ 2 * 4 + j ]
-				+ a [ i * 4 + 3 ] * b [ 3 * 4 + j ];
+void R_MatrixMultiply( const float a[16], const float b[16], float out[16] ) {
+	int cpuid = idLib::sys->GetProcessorId();
+
+	/* fast path use sse */
+	if ( cpuid & CPUID_SSE ) {
+		__m128 B0x = _mm_loadu_ps( &b[0] );
```
DanielGibson, line 770: I think this won't even compile on non-x86 (or x86_64) machines, so it must be behind some kind of `#if` or `#ifdef`

neo/renderer/tr_main.cpp
```diff
@@ -599,16 +585,16 @@ bool R_RadiusCullLocalBox( const idBounds &bounds, const float modelMatrix[16],
 
 	R_LocalPointToGlobal( modelMatrix, localOrigin, worldOrigin );
 
-	worldRadius = (bounds[0] - localOrigin).Length();	// FIXME: won't be correct for scaled objects
+	worldRadius = ( bounds[0] - localOrigin ).Length();	// FIXME: won't be correct for scaled objects
 
 	for ( i = 0 ; i < numPlanes ; i++ ) {
 		frust = planes + i;
 		d = frust->Distance( worldOrigin );
+
```
DanielGibson, line 593: please don't commit all these unrelated whitespace changes. use `git gui` or similar to only commit the relevant changes

revelator: sorry been hammering away at getting it to compile on gcc / clang.
mingw-w64 builds work now as well as windows clang builds.

there was one sorry bugger with current mingw-w64-abi the __cpuid function now uses the same syntax as msvc's so i had to change the #if defined(__GNUC__) && (defined(__i386__) || defined (__x86_64__)) guard into one specific for unix  so the mingw builds dont use that one.

well atleast it was informal i reckon id make a new repo after getting the last build bugs out and start slowly adding in the changes, that way we can avoid all the whitespace changes from me mucking around in the code.

neo/renderer/tr_main.cpp
revelator: should be fixed now havent comitted the new changes yet though (different machine so i need to setup a few things).

it does compile with mingw-w64 and clang now but i noticed i forgot to change the __SSE2__ macro to __SSE__ which caused it to fail.

might need another guard for unix / mac lets see.

revelator:
i added some test builds made with mingw-w64 and clang here ->
https://sourceforge.net/projects/cbadvanced/files/dhewm3-clang64_2.zip/download
https://sourceforge.net/projects/cbadvanced/files/dhewm3-clang32_2.zip/download
https://sourceforge.net/projects/cbadvanced/files/dhewm3-mingw64_2.zip/download
https://sourceforge.net/projects/cbadvanced/files/dhewm3-mingw32_2.zip/download

loads of dependencies due to msys2 wanting to have everything linked to dll's xD

neo/renderer/tr_main.cpp
revelator: for now lets use it as a draft to iron out build bugs on different arch's, bit easier than having multiple operating systems 
:) when it is all up and running ill retire the fork and start from a fresh one adding only the changes nessesary without the formatting. also gives you the option to chime in before we make changes that may make or break something.

revelator: ok seems to build on ubuntu now atleast in the one in wsl.
running it on wsl does not seem to work though the game window opens it crashes shortly after with a crash in imgui (not sure wsl is the best testing environment).

i packed up the binaries here https://sourceforge.net/projects/cbadvanced/files/Game%20Projects/dhewm3-ubuntu.7z/download

could you try them out on a real linux box maybe ?.

revelator: https://sourceforge.net/projects/cbadvanced/files/Game%20Projects/dhewm3-git.tar.zst/download

my current source code ignore whitespace it is not ment to be merged as is only for testing if it works :).

https://sourceforge.net/projects/cbadvanced/files/Game%20Projects/dhewm3.tar.zst/download

i forgot linux sets some flags that are probably lost when zipping it with a windows tool so heres a tarball of the game executables made in ubuntu, i used the linux dist dir in the source as install target (not sure if correct ?).

revelator:
ok most whitespace changes are now gone in my local build tree.

one question though __cpuid on mac does something odd to the cpu registers i guess that was why your version exchanged the ebx value for esi, i had to revert it partially because even though the version from darkmod worked in linux it broke when trying to compile it for mac.

i hope to god once i uploaded the change to cpu.cpp that all build error's are gone so i can start seperating the different changes in a new fork.

neo/renderer/tr_main.cpp
```diff
@@ -473,9 +474,9 @@ void R_AxisToModelMatrix( const idMat3 &axis, const idVec3 &origin, float modelM
     modelMatrix[15] = 1;
 }
 
-// FIXME: these assume no skewing or scaling transforms
+// FIXME: these assume no skewing or scaling transforms (disabled SSE intrinsics on MAC ppc)
 void R_LocalPointToGlobal( const float modelMatrix[16], const idVec3 &in, idVec3 &out ) {
-#if !( defined(MACOS_X) || defined(__APPLE__) )
+#if !( defined(__ppc__) || defined(MACOS_X) || defined(__APPLE__) )
```
DanielGibson, line 479: I don't think this is a very good check for this purpose. It doesn't do what you want on
* x86 Macs (there it should be enabled)
* non-x86 CPUs that aren't PPC (unless they happen to run OSX)

Can't you just use `#ifdef __SSE__`?

neo/idlib/math/Simd_AVX.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+//===============================================================
+//
+//  AVX implementation of idSIMDProcessor
+//
+//===============================================================
+
+// Revelator: these work whether in gcc clang or msvc in x86 or x64 (no inline assembly used)
+#if defined(_MSC_VER) && ( defined(_M_X64) || defined(_M_IX86) ) || \
+    defined(__GNUC__) && ( defined(__i386__) || defined (__x86_64__) ) && defined(__AVX__)
```
DanielGibson, line 40: With GCC, this currently disabled the code because __AVX__ is only defined if `-mavx` is used - which would have to be set in CMakeLists.txt, only for this file

DanielGibson: also, I guess `__AVX__` is only set if `__i386__` or `__x86_64__` is set, so those checks aren't necessary

neo/renderer/tr_main.cpp
revelator: yeah i newer coded for MAC before so was a bit unsure how to proceed :) ill see if just using __SSE__ helps.

neo/idlib/math/Simd_AVX.cpp
revelator: yeah i been working on some cmake changes to probe the users cpu and set the correct flags for gnu, still not ready but ill post the change when it is.

but is it only avx that needs a flag set or do we also need it for every SSE type ? like -sse -sse2 -sse3 -ssse3 -mavx -mavx2 etc. if so that would suck.

DanielGibson: hmm maybe this could also be handled in the source files instead of CMake, like
```c++
// add this at the beginning of the source file, under the big comment block with the license
#ifdef __GNUC__
#pragma GCC target (avx)
#endif
```

https://gcc.gnu.org/onlinedocs/gcc/x86-Function-Attributes.html#index-target-function-attribute-5 has a list of valid target arguments

DanielGibson: apparently for clang it's slighly more complicated, see https://stackoverflow.com/a/56395854

so maybe it must be something like
```c++
#ifdef __clang__
#pragma clang attribute push (__attribute__((target("avx"))), apply_to=function)
#elif defined __GNUC__
#pragma GCC target ("avx")
#endif
```

though it's possible that in practice the older extensions (>SSE2) must also be listed, if the code uses intrinsics from them as well, that might look like
`#pragma clang attribute push (__attribute__((target("sse3", "ssse3", "avx"))), apply_to=function)` or `#pragma GCC target ("sse3", "ssse3", "avx")` (just an example if SSE3, SSSE3 and AVX are needed)

revelator: hmm interresting i had no idea gcc supported setting intrinsics via pragmas but cool if it does :) 
guess we need to ask around the clang user base for specifics luckily i know one of the devs martin storsjoe he might have some insight on this.

revelator: hmm yes for clang this does seem the right approach allthough we also need to pop the stack afterwards.
for gcc #pragma GCC target("avx2,fma") should also work for avx2. can even set tune to the lowest cpu supporting it for more fine control like eg. tune=haswell maybe ?. also seems the pragmas need to come after the intrinsics header otherwise they break.

darkmod has made a lot of changes to intrinsics for one they ditched all the older simd code using the portable intrinsics instead.

revelator:
```
//===============================================================
//
//  AVX2 implementation of idSIMDProcessor
//
//===============================================================

// Revelator: these work whether in gcc clang or msvc in x86 or x64 (no inline assembly used)
#include <immintrin.h>

// must come after immintrin.h
#if defined(__i386__) || defined (__x86_64__)
#ifdef __clang__
#pragma clang attribute push (__attribute__((target("avx2,fma"))), apply_to=function)
#elif defined __GNUC__
#pragma GCC target ("avx2,fma")
#endif
#endif // x86 or x86_64

#include "idlib/geometry/DrawVert.h"
#include "idlib/geometry/JointTransform.h"
#include "idlib/math/Vector.h"
#include "idlib/math/Plane.h"

/*
============
idSIMD_AVX2::GetName
============
*/
const char *idSIMD_AVX2::GetName( void ) const {
    return "MMX & SSE & SSE2 & SSE3 & AVX & AVX2";
}

/*
============
idSIMD_AVX2::CullByFrustum
============
*/
void VPCALL idSIMD_AVX2::CullByFrustum( idDrawVert *verts, const int numVerts, const idPlane frustum[6], byte *pointCull, float epsilon ) {
    const __m256 fA = _mm256_set_ps( 0, 0, frustum[5][0], frustum[4][0], frustum[3][0], frustum[2][0], frustum[1][0], frustum[0][0] );
    const __m256 fB = _mm256_set_ps( 0, 0, frustum[5][1], frustum[4][1], frustum[3][1], frustum[2][1], frustum[1][1], frustum[0][1] );
    const __m256 fC = _mm256_set_ps( 0, 0, frustum[5][2], frustum[4][2], frustum[3][2], frustum[2][2], frustum[1][2], frustum[0][2] );
    const __m256 fD = _mm256_set_ps( 0, 0, frustum[5][3], frustum[4][3], frustum[3][3], frustum[2][3], frustum[1][3], frustum[0][3] );
    const __m256 eps = _mm256_set1_ps( epsilon );
    const byte mask6 = ( 1 << 6 ) - 1;

    for ( int j = 0; j < numVerts; j++ ) {
        idVec3 &vec = verts[j].xyz;
        __m256 vX = _mm256_set1_ps( vec.x );
        __m256 vY = _mm256_set1_ps( vec.y );
        __m256 vZ = _mm256_set1_ps( vec.z );
        __m256 d =  _mm256_fmadd_ps( fA, vX,
                    _mm256_fmadd_ps( fB, vY,
                    _mm256_fmadd_ps( fC, vZ, fD ) ) );
        int mask_lo = _mm256_movemask_ps( _mm256_cmp_ps( d, eps, _CMP_LT_OQ ) );
        pointCull[j] = ( byte )( mask_lo & mask6 ); // gcc compiler warning
    }
    _mm256_zeroupper();
}

/*
============
idSIMD_AVX2::CullByFrustum2
============
*/
void VPCALL idSIMD_AVX2::CullByFrustum2( idDrawVert *verts, const int numVerts, const idPlane frustum[6], unsigned short *pointCull, float epsilon ) {
    const __m256 fA = _mm256_set_ps( 0, 0, frustum[5][0], frustum[4][0], frustum[3][0], frustum[2][0], frustum[1][0], frustum[0][0] );
    const __m256 fB = _mm256_set_ps( 0, 0, frustum[5][1], frustum[4][1], frustum[3][1], frustum[2][1], frustum[1][1], frustum[0][1] );
    const __m256 fC = _mm256_set_ps( 0, 0, frustum[5][2], frustum[4][2], frustum[3][2], frustum[2][2], frustum[1][2], frustum[0][2] );
    const __m256 fD = _mm256_set_ps( 0, 0, frustum[5][3], frustum[4][3], frustum[3][3], frustum[2][3], frustum[1][3], frustum[0][3] );
    const __m256 eps = _mm256_set1_ps( epsilon );
    static const __m256 epsM = _mm256_set1_ps( -epsilon );
    const short mask6 = ( 1 << 6 ) - 1;

    for ( int j = 0; j < numVerts; j++ ) {
        idVec3 &vec = verts[j].xyz;
        __m256 vX = _mm256_set1_ps( vec.x );
        __m256 vY = _mm256_set1_ps( vec.y );
        __m256 vZ = _mm256_set1_ps( vec.z );
        __m256 d =  _mm256_fmadd_ps( fA, vX,
                    _mm256_fmadd_ps( fB, vY,
                    _mm256_fmadd_ps( fC, vZ, fD ) ) );
        int mask_lo = _mm256_movemask_ps( _mm256_cmp_ps( d, eps, _CMP_LT_OQ ) );
        int mask_hi = _mm256_movemask_ps( _mm256_cmp_ps( d, eps, _CMP_GT_OQ ) );
        pointCull[j] = ( unsigned short )( ( mask_lo & mask6 ) | ( mask_hi & mask6 ) ) << 6; // gcc compiler warning
    }
    _mm256_zeroupper();
}

#if defined(__i386__) || defined (__x86_64__)
#ifdef __clang__
#pragma clang attribute pop
#endif
#endif
```

something like this maybe ?

revelator: seems we also need #pragma GCC push_options before the #pragma GCC target string and subsequent stack pop as well for gcc.

revelator:
```
#if defined(__i386__) || defined (__x86_64__)
#ifdef __clang__
#pragma clang attribute push (__attribute__((target("avx,avx2,fma,arch=haswell"))), apply_to=function)
#elif defined __GNUC__
#pragma GCC push_options
#pragma GCC target ("avx,avx2,fma,arch=haswell")
#endif
#endif // x86 or x86_64
```
there we go avx2 depends on avx and fma but as far as i can see not on any of the older sse mechanics.

DanielGibson: could be, TBH I wasn't even sure if this works at all, was just an idea because I know you can locally enable/disable optimization like this

my assumption about pushing/popping was that it doesn't matter because we want those options to be used for the whole source file - but could be that compilers insist on it anyway, IDK

DanielGibson: BTW: `// must come after immintrin.h` why? I'd think that the intrinsics header also needs to know we're now targeting AVX instead of just SSE2

DanielGibson: Another thing: I don't think `arch=haswell` should be set.
We don't want the generated code to be optimized for haswell, or to even use instructions that are only available on haswell but not some other CPUs that might support avx2 (or whatever) anyway (for example by using instructions from other CPU extensions that only Intel supports, but AMD doesn't).

The GCC manpage says:
>   `-march=cpu-type`
>       Generate instructions for the machine type cpu-type.  In contrast to `-mtune=cpu-type`, which merely tunes the generated code for the specified cpu-type, `-march=cpu-type` allows GCC to **generate code that may _not run at all_ on processors other than the one indicated**.  Specifying `-march=cpu-type` implies `-mtune=cpu-type`, except where noted otherwise.

revelator: oh was not ment for haswell optimization just to make sure we dont use the code on a cpu that does not support it, otherwise it will surely crash. or should we use -mtune ?.

the other idea sounds great :)

revelator: oh fotgot the reason it must come after the intrinsic header is because otherwise the pragma fails, this was noticed by someone doing something similar so i took my cue from that :)

DanielGibson: >  just to make sure we dont use the code on a cpu that does not support it, otherwise it will surely crash. 

yes, and that's exactly why -march should **not** be set :-p

> or should we use -mtune ?.

no, just use "avx2,fma" if the intrinsics code in that file needs AVX2 and FMA

Then the compiler should only use instructions that are either available in x86 with SSE and SSE2, **or** the extensions specified in the pragma (AVX2 and FMA in this case).

Note: x86_64 aka x64 always supports SSE and SSE2, and when building 32bit x86 code with MSVC it nowadays also defaults to supporting SSE2 (though that's not relevant for these GCC/clang-specific macros). Currently when using GCC/clang to build for 32bit x86 dhewm3 does *not* use SSE(2), but I plan to change that in CMake.

DanielGibson: > oh forgot the reason it must come after the intrinsic header is because otherwise the pragma fails, this was noticed by someone doing something similar

We should try this ourselves to see if that's actually (still) true.

revelator: im game :)

revelator: avx2 needs avx as well according to darkmod so avx,avx2,fma it seems.

DanielGibson: that sounds plausible, use that then :)

(no idea if `-mavx2` implies `-mavx` - it would make sense, I don't think there are any CPUs implementing AVX2 but not AVX, but IDK, the manpage doesn't mention it)

revelator: btw just for giggles and to see if my sdl wrappers in cpu.cpp worked i tried a build with SDL1.5 and oh boy...
needs some missing gl definitions not supplied by SDL1 and there was this one in SDL_win32_main.c

```
	/* Load SDL dynamic link library */
	if ( SDL_Init(SDL_INIT_NOPARACHUTE) < 0 ) {
		ShowError("WinMain() error", SDL_GetError());
		return(FALSE);
	}
	//atexit(cleanup_output); // uh referencing a static function in another source file ?
	atexit(cleanup);
```
note the one i commented.

so after pulling in all the gl definitions and commenting that one it runs but the mouse is all wonky (sometimes dissapearing) possible oversight ? or should that function not have been static maybe ?.

revelator: ok ill go with darkmods suggestion :) shouldnt hurt at least.

revelator: agree i just thought you keept the SDL1 code for posterity or possibly some ancient machine ;)
guess it should be removed at some point.

hmm SDL_win32_main.cpp also only seems to be pulled in if you specifically request SDL1.

revelator: i foresee one problem atleast with my hybrid glsl/arb renderer since SDL3 no longer supports the old hardware gamma we would need shader code for the glsl interactions for gamma as well when moving on to SDL3.

but lets get the intrinsics working first :)

DanielGibson: > hmm SDL_win32_main.cpp also only seems to be pulled in if you specifically request SDL1.

oh right, now I remember.
yes, that `atexit(cleanup_output);` line can be deleted, `cleanup_output()` has been moved to win_main.cpp and `atexit(cleanup_output);` is now called from `SDL_main()` in win_main.cpp (for all SDL versions)

DanielGibson: I'm not sure I want GLSL at all at this point, but if it's added it definitely needs to support shader gamma.

revelator: normally would not be a problem but because of the way it is loaded directly in RB_ARB2_DrawInteraction it causes problems for my renderer because it does the housekeeping itself if told this one goes to arb then it goes to the arb side and vice versa but since it is loaded directly in the interaction renderer it bonks out if glsl is selected.
as an example i can run sikkmod with glsl interactions and still use all the arb effects :)

DanielGibson:
I created a short testcase for D3_PUSH_X86_TARGETS(): https://gist.github.com/DanielGibson/4e35947c548523e2ac42047a76df9239

it needs to be implemented like in that file, the code I suggested earlier didn't work

neo/sys/cpu.cpp
```diff
@@ -28,205 +28,414 @@ If you have questions concerning this license or the applicable additional terms
...
+    processorString[3] = pstring[7];
+    processorString[4] = pstring[12];
+    processorString[5] = pstring[13];
+    processorString[6] = pstring[14];
+    processorString[7] = pstring[15];
+    processorString[8] = pstring[8];
+    processorString[9] = pstring[9];
+    processorString[10] = pstring[10];
+    processorString[11] = pstring[11];
+    processorString[12] = 0;
```
DanielGibson, line 107: what's the point of copying pstring to processorString?

revelator: not sure either it's from darkmod stgatilov might know ?

revelator: null terminating pstring ? but there are better ways.

revelator: sorry had to get some shuteye it was allmost 2.00 in the night here :)
not a young man anymore sadly xD

revelator: for terminating some say 0 is sufficient if using C strings though i hear quite a lot prefer the special '\0' character.

tbh in a c++ project std::string would probably have been better.

DanielGibson: not sure how a std::string would help here, but for 0-termination `pstring[12] = 0;` should suffice?

DanielGibson: Looks like this code is directly from the original Doom3 GPL release: https://github.com/TTimo/doom3.gpl/blob/aaa855815ab484d5bd095f347163194ac569abcc/neo/sys/win32/win_cpu.cpp#L199-L223

still no idea why they did that, and I also think that it doesn't really matter if the users CPU is from AMD (original Doom3 used that to decide what clockrate a CPU needs to qualify for medium/high quality, but dhwem3 doesn't do that. https://github.com/TTimo/doom3.gpl/blob/aaa855815ab484d5bd095f347163194ac569abcc/neo/framework/Common.cpp#L2746-L2749)

revelator: hmm could be right i admit i did not look at the original code release.
not sure myself why it was done like that myself.

yeah it probably makes less sense with the check today seing as ryzen supports anything intel does, not even sure if needed at the time of 3DNow since that was more like an extension of the MMX model.
probably could be used for determining unsupported cpu types but since there are little of those left and the ones left dont support playing doom3 anyway it might be prudent to remove it unless users actually want to look at cpu information in the doom3 console :)

FriskTheFallenHuman:
Any news on this?

Issue #544, opened 2024-01-10
Nothing happens when I launch the application


petran79:
I copied all the files in base and d3xp folder 
.bat files lists dhewm3.exe +set fs_game d3le +set fs_game_base d3xp

However nothing happens when I click on the file, not even an error message,
I am using WIndows 10. Do not know if it has to do with switching to a gsync monitor 

Original Doom 3 on Steam do not run either. Only FHDOOM runs fine

petran79:
I do not have any antivirus checker on Windows 10 and even disabled Windows Defender via Defender Control utility. Still dhewm does not run and no message appears. Even installing the VC+ does not solve the issue.

DanielGibson:
Is this still a problem?
I have no idea why that should happen, except for antivirus running amok or maybe an incorrect GPU driver.

Does only D3LE not work or does the basegame (if you just run `dhewm3.exe`) not work either?

petran79:
same issue. Clean installed the game and copied the dhewm contents to base game folder. Base game and roe run fine from Steam and also outside Steam , yet dhewm3. exe refuses to run from doom 3 base game folder
Even after various Windows 10 and gpu driver updates over the monthsl

DanielGibson:
and you don't get any error message?
does anything turn up in your antivirus logs?

DanielGibson:
What GPU do you have anyway?

petran79:
no error message at all, not even from command line instead of windows Explorer. Everything blank.
Only antivirus is windows defender which is disabled. 

I use an old gtx 1080. Cpu is also an old i7 4790k. The thing is if Windows had blocked it (dep) , at least a blue message would let me know. Not even this appears.

Other doom3 source ports like cstdoom3 run fine.
tried moving it to c: too but no difference.

DanielGibson:
Do you have the latest dhewm3 version (1.5.4)?
Do you have Visual Studio installed? If so, what happens when you build dhewm3 yourrself and run it in the debugger?

If not, try installing WinDbg (https://apps.microsoft.com/detail/9pgjgd53tn86) and running dhewm3 in it, hopefully WinDbg will provide any information why it doesn't run (or maybe why it shuts down so fast it looks like it doesn't run)

petran79:
I installed it. 
When I tried launching it gives me this error


************* Preparing the environment for Debugger Extensions Gallery repositories **************
   ExtensionRepository : Implicit
   UseExperimentalFeatureForNugetShare : true
[...]

PR #642, opened 2025-01-18
Fixes for older VS versions, make sure dhewm3log.txt was created on Windows

DanielGibson:
#544 has shown that dhewm3 will silently fail to start on Windows if dhewm3log.txt can't be created (for example because Documents/My Games/dhewm3/ can't be written to) and that debugging that is a major PITA.
Now a Windows MessageBox with an error message is shown.

Also fixed an issue that prevented VS2017 from compiling dhewm3 with ImGui enabled.


neo/sys/win32/win_main.cpp
```diff
@@ -1142,6 +1160,12 @@ static void redirect_output(void)
 		newfp = fopen(stderrPath, TEXT("w"));
 		if ( newfp ) {
 			*stderr = *newfp;
+		} else {
+			char msg[2048];
+			D3_snprintfC99( msg, sizeof(msg), "Failed to create '%s',\n error number is %d (%s)\nIs Documents/My Games/dhewm3/ write protected?",
+				stdoutPath, errno, strerror(errno) );
```
turol, line 1166: Copy-paste error, wrong variable.

DanielGibson: good find!

PR #643, opened 2025-01-19
Support lights "nospecular" parm (if explicitly enabled)

DanielGibson:
based on https://github.com/dhewm/dhewm3/pull/254

The "nospecular" parm will only be used if either
r_supportNoSpecular is set to 1
or r_supportNoSpecular is set to -1 (the default) and the maps spawnargs contain "allow_nospecular" "1"

Drawback: I think the current code doesn't work with (time)demos, as I *think* that I can't access the maps spawnargs when playing one

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -23,6 +23,11 @@ Note: Numbers starting with a "#" like #330 refer to the bugreport with that num
 * Updated Dear ImGui to 1.91.4
 * Fix scaling of Grabber cursor in Resurrection of Evil in non-4:3 resolutions (#637)
 * Add `com_disableAutoSaves` CVar: If set to `1`, Autosaves (when starting a level) are disabled (#620)
+* Add support for "nospecular" parm of lights, enabled by setting `"allow_nospecular" "1"` in a maps
+  worldspawn, or by setting the `r_allowNoSpecular` CVar to `1`.  
+  Note that this required changing the format of demos. dhewm3 can still play old demos, but ones
+  recorded with current dhewm3 are not compatible with older dhewm3 versions, original Doom3 or
+  other source ports (unless they do the same change).
 
 1.5.4 (2024-08-03)
 ------------------------------------------------------------------------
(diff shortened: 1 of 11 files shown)


DanielGibson:
This needs testing, which implies a custom testmap (that someone - not me - has to create).
I won't merge this if it's not tested!

DanielGibson:
@FriskTheFallenHuman was so friendly to create a testmap for this: 
[nospecular_test.zip](https://github.com/user-attachments/files/18470824/nospecular_test.zip)

This seems to work as expected.

There are two things that need a bit more investigation:
1. (time)demos
2. D3Radiant - @FriskTheFallenHuman said that this worked in the level editor, but I don't understand how/why/when. Ideally nospecular should only work there if `"allow_nospecular" "1"` is set in the worldspawn (and then ideally it should work immediately without reloading the map), but I didn't add any code to enable that so I'm surprised it apparently works at all..

DanielGibson:
The radiant part of the problem is solved, it now works reliably and updates `tr.allowNoSpecular` whenever `allow_nospecular` is set/changed in the worldspawn, and when a map is loaded.

Still not sure what to do about the timedemos, though.

FriskTheFallenHuman:
> The radiant part of the problem is solved, it now works reliably and updates `tr.allowNoSpecular` whenever `allow_nospecular` is set/changed in the worldspawn, and when a map is loaded.
> 
> Still not sure what to do about the timedemos, though. I could just change the demoformat to save the current `tr.allowNoSpecular` value when recording and set that when playing... Is that bad? dhewm3 could still play old demos, but new demos would be incompatible with older versions of dhewm3 or Vanilla Doom3 or other source ports.

perhaps implementing a new timedemo format may not be that bad, has far i know it shouldn't be a problem since timedemos where not that common unlike standard demos iirc

DanielGibson:
I mean standard demos, timedemos just play standard demos at high speed

DanielGibson:
I changed it now.
Still not sure if it matters - is recording demos and redistributing them something that's (still) commonly done?

If this really turns into a problem I could create a CVar that allows choosing the demo recording format

DanielGibson:
no idea why github doesn't detect that, but I rebased and merged the branch manually

PR #649, opened 2025-01-31
Feature: Porting MFC tools

ashalkhakov:
This adds editors ported from MFC and builds on top of [imgui-docking](https://github.com/DanielGibson/dhewm3/commits/imgui-docking/).

Currently ported tools:

* PDA editor

image

* Particle editor (WIP, mostly done)

image

* Decl browser (WIP)

image

* Script Editor (with working autosuggestions and tooltips)

image

### DIFF
--- a/neo/tools/imgui/materialeditor/ConsoleView.cpp
+++ b/neo/tools/imgui/materialeditor/ConsoleView.cpp
@@ -176,6 +176,9 @@ void ConsoleView::Draw( const ImVec2 &pos, const ImVec2 &size ) {
 	if ( selectEditInput ) {
 		selectEditInput = false;
 	}
+	if ( ImGui::IsWindowFocused() ) {
+		MaterialEditorSetActiveWindow( ME_WINDOW_CONSOLE );
+	}
 }
 
 /**
--- a/neo/tools/imgui/materialeditor/MEMainFrame.cpp
+++ b/neo/tools/imgui/materialeditor/MEMainFrame.cpp
@@ -103,6 +103,8 @@ MEMainFrame::MEMainFrame() {
 
 	searchData.searched = false;
 
+	activeWindow = ME_WINDOW_NONE;
+
 	options.Load();
 }
 
@@ -326,8 +328,7 @@ void MEMainFrame::Draw() {
 
 			ImGui::Separator();
 
-			if( ImGui::MenuItem( "Exit", "Ctrl+W" ) )
-			{
+			if( ImGui::MenuItem( "Exit", "Ctrl+W" ) || ImGui::Shortcut( ImGuiMod_Ctrl | ImGuiKey_W ) ) {
 				OnFileExit();
 			}
 
@@ -336,8 +337,7 @@ void MEMainFrame::Draw() {
 		if ( ImGui::BeginMenu( "Edit" ) ) {
 			ImGui::BeginDisabled( !IsEditUndoEnabled() );
 			ImGui::SetNextItemShortcut( ImGuiMod_Ctrl | ImGuiKey_Z );
-			ImGui::MenuItem( "Undo", "Ctrl+Z" );
-			if ( ImGui::Shortcut( ImGuiMod_Ctrl | ImGuiKey_Z ) ) { 
+			if ( ImGui::MenuItem( "Undo", "Ctrl+Z" ) || ImGui::Shortcut( ImGuiMod_Ctrl | ImGuiKey_Z ) ) {
 				OnEditUndo();
 			}
 			ImGui::EndDisabled();
@@ -354,37 +354,32 @@ void MEMainFrame::Draw() {
 
 			ImGui::BeginDisabled( !IsEditCutEnabled() );
 			ImGui::SetNextItemShortcut( ImGuiMod_Ctrl | ImGuiKey_X );
-			ImGui::MenuItem( "Cut", "Ctrl+X" );
-			if ( ImGui::Shortcut( ImGuiMod_Ctrl | ImGuiKey_X ) ) {
+			if ( ImGui::MenuItem( "Cut", "Ctrl+X" ) || ImGui::Shortcut( ImGuiMod_Ctrl | ImGuiKey_X ) ) {
 				OnEditCut();
 			}
 			ImGui::EndDisabled();
 
 			ImGui::BeginDisabled( !IsEditCopyEnabled() );
 			ImGui::SetNextItemShortcut( ImGuiMod_Ctrl | ImGuiKey_C );
-			ImGui::MenuItem( "Copy", "Ctrl+C" );
-			if ( ImGui::Shortcut( ImGuiMod_Ctrl | ImGuiKey_C ) ) {
+			if ( ImGui::MenuItem( "Copy", "Ctrl+C" ) || ImGui::Shortcut( ImGuiMod_Ctrl | ImGuiKey_C ) ) {
 				OnEditCopy();
 			}
 			ImGui::EndDisabled();
 			ImGui::BeginDisabled( !IsEditPasteEnabled() );
 			ImGui::SetNextItemShortcut( ImGuiMod_Ctrl | ImGuiKey_V );
-			ImGui::MenuItem( "Paste", "Ctrl+V" );
-			if ( ImGui::Shortcut( ImGuiMod_Ctrl | ImGuiKey_V ) ) {
+			if ( ImGui::MenuItem( "Paste", "Ctrl+V" ) || ImGui::Shortcut( ImGuiMod_Ctrl | ImGuiKey_V ) ) {
 				OnEditPaste();
 			}
 			ImGui::EndDisabled();
 			ImGui::BeginDisabled( !IsEditDeleteEnabled() );
 			ImGui::SetNextItemShortcut( ImGuiKey_Delete );
-			ImGui::MenuItem( "Delete", "Del" );
-			if ( ImGui::Shortcut( ImGuiKey_Delete ) ) {
+			if ( ImGui::MenuItem( "Delete", "Del" ) || ImGui::Shortcut( ImGuiKey_Delete ) ) {
 				OnEditDelete();
 			}
 			ImGui::EndDisabled();
 			ImGui::BeginDisabled( !IsEditRenameEnabled() );
 			ImGui::SetNextItemShortcut( ImGuiKey_F2 );
-			ImGui::MenuItem( "Rename", "F2" );
-			if ( ImGui::Shortcut( ImGuiKey_F2 ) ) {
+			if ( ImGui::MenuItem( "Rename", "F2" ) || ImGui::Shortcut( ImGuiKey_F2 ) ) {
 				OnEditRename();
 			}
 			ImGui::EndDisabled();
@@ -392,7 +387,7 @@ void MEMainFrame::Draw() {
 			ImGui::Separator();
 
 			ImGui::SetNextItemShortcut( ImGuiMod_Ctrl | ImGuiKey_F );
-			if ( ImGui::MenuItem( "Find", "Ctrl+F" ) ) {
+			if ( ImGui::MenuItem( "Find", "Ctrl+F" ) || ImGui::Shortcut( ImGuiMod_Ctrl | ImGuiKey_F ) ) {
 				OnEditFind();
 			}
 
@@ -442,10 +437,20 @@ void MEMainFrame::Draw() {
 		ImGui::EndMenuBar();
 	}
 
+	// debug
+	/*ImGui::Text( "Focus: %s",
+		activeWindow == ME_WINDOW_NONE ? "None"
+		: activeWindow == ME_WINDOW_TREE ? "Tree"
+		: activeWindow == ME_WINDOW_TEXT_EDIT ? "Text Edit"
+		: activeWindow == ME_WINDOW_PROP ? "Prop"
+		: activeWindow == ME_WINDOW_PREVIEW ? "Preview"
+		: activeWindow == ME_WINDOW_PREVIEW_PROP ? "Preview Prop"
+		: activeWindow == ME_WINDOW_STAGE ? "Stage"
+		: "Console" );*/
+
 	if ( ImGui::BeginTabBar( "MainTabBar" ) ) {
 
-		if (ImGui::BeginTabItem( "Editor" ) )
-        {
+		if ( ImGui::BeginTabItem( "Editor" ) ) {
 			float splitterButtonWidthOrHeight = 8.0f;
 			ImVec2 windowSize = ImVec2( Max( editSplitterWidth, previewSplitterWidth ) + splitterButtonWidthOrHeight, editSplitterHeight + previewSplitterHeight + splitterButtonWidthOrHeight * 2.0f );
 
@@ -477,7 +482,7 @@ void MEMainFrame::Draw() {
 					m_find->Draw( ImVec2( previewSplitterPos, 0 ), ImVec2( editSplitterWidth, editSplitterHeight ) );
 				}
 
-				ImGui::InvisibleButton("editRowHSplitter", ImVec2(-1, splitterButtonWidthOrHeight));
+				ImGui::InvisibleButton( "editRowHSplitter", ImVec2( -1, splitterButtonWidthOrHeight ) );
 				if ( ImGui::IsItemActive() ) {
 					editSplitterHeight += ImGui::GetIO().MouseDelta.y;
 				}
@@ -519,6 +524,10 @@ void MEMainFrame::Draw() {
 		}
 		ImGui::EndTabBar();
 	}
+}
+
+void MEMainFrame::SetActiveWindow( MaterialEditorWindow_t window ) {
+	activeWindow = window;
 }
 
 /**
@@ -638,69 +647,89 @@ bool MEMainFrame::IsApplyAllEnabled() {
 * Performs a cut operation on the selected material.
 */
 void MEMainFrame::OnEditCut() {
-	/*CWnd* focus = GetFocus();
-	if(focus) {
-		if (focus->IsKindOf(RUNTIME_CLASS(MaterialTreeView))) {
-			m_materialTreeView->OnCut();
-		}
-	}*/
+	switch ( activeWindow ) {
+	case ME_WINDOW_TREE:
+		m_materialTreeView->OnCut();
+		break;
+	case ME_WINDOW_TEXT_EDIT:
+		m_materialEditView->OnCut();
+		break;
+	default:
+		break;
+	}
 }
 
 /**
 * Performs a copy operation on the selected material or stage.
 */
 void MEMainFrame::OnEditCopy() {
-	/*CWnd* focus = GetFocus();
-	if(focus) {
-		if (focus->IsKindOf(RUNTIME_CLASS(StageView))) {
-			m_stageView->OnCopy();
-		} else if (focus->IsKindOf(RUNTIME_CLASS(MaterialTreeView))) {
-			m_materialTreeView->OnCopy();
-		}
-	}*/
+	switch ( activeWindow ) {
+	case ME_WINDOW_STAGE:
+		m_stageView->OnCopy();
+		break;
+	case ME_WINDOW_TREE:
+		m_materialTreeView->OnCopy();
+		break;
+	case ME_WINDOW_TEXT_EDIT:
+		m_materialEditView->OnCopy();
+		break;
+	default:
+		break;
+	}
 }
 
 /**
 * Performs a paste operation on the selected material or stage.
 */
 void MEMainFrame::OnEditPaste() {
-	/*CWnd* focus = GetFocus();
-
-	if(focus) {
-		if (focus->IsKindOf(RUNTIME_CLASS(StageView))) {
-			m_stageView->OnPaste();
-		} else if (focus->IsKindOf(RUNTIME_CLASS(MaterialTreeView))) {
-			m_materialTreeView->OnPaste();
-		}
-	}*/
+	switch ( activeWindow ) {
+	case ME_WINDOW_STAGE:
+		m_stageView->OnPaste();
+		break;
+	case ME_WINDOW_TREE:
+		m_materialTreeView->OnPaste();
+		break;
+	case ME_WINDOW_TEXT_EDIT:
+		m_materialEditView->OnPaste();
+		break;
+	default:
+		break;
+	}
 }
 
 /**
 * Performs a delete operation on the selected material or stage.
 */
 void MEMainFrame::OnEditDelete() {
-	/*CWnd* focus = GetFocus();
-	if(focus) {
-		if (focus->IsKindOf(RUNTIME_CLASS(StageView))) {
-			m_stageView->OnDeleteStage();
-		} else if (focus->IsKindOf(RUNTIME_CLASS(MaterialTreeView))) {
-			m_materialTreeView->OnDeleteMaterial();
-		}
-	}*/
+	switch ( activeWindow ) {
+	case ME_WINDOW_STAGE:
+		m_stageView->OnDeleteStage();
+		return;
+	case ME_WINDOW_TREE:
+		m_materialTreeView->OnDeleteMaterial();
+		break;
+	case ME_WINDOW_TEXT_EDIT:
+		m_materialEditView->OnDelete();
+		break;
+	default:
+		break;
+	}
 }
 
 /**
 * Performs a rename operation on the selected material or stage.
 */
 void MEMainFrame::OnEditRename() {
-	/*CWnd* focus = GetFocus();
-	if(focus) {
-		if (focus->IsKindOf(RUNTIME_CLASS(StageView))) {
-			m_stageView->OnRenameStage();
-		} else if (focus->IsKindOf(RUNTIME_CLASS(MaterialTreeView))) {
-			m_materialTreeView->OnRenameMaterial();
-		}
-	}*/
+	switch ( activeWindow ) {
+	case ME_WINDOW_STAGE:
+		m_stageView->OnRenameStage();
+		break;
+	case ME_WINDOW_TREE:
+		m_materialTreeView->OnRenameMaterial();
+		break;
+	default:
+		break;
+	}
 }
 
 
@@ -764,12 +793,21 @@ void MEMainFrame::OnEditFindNext() {
 */
 void MEMainFrame::OnEditUndo() {
 	//Check for undo operation on special windows
-	/*if ( ImGui::GetID( "Text" ) == ImGui::GetActiveID() ) {
+	switch ( activeWindow ) {
+	case ME_WINDOW_TEXT_EDIT:
 		m_materialEditView->m_textView.Undo();
-		return;
-	}*/
-
-	materialDocManager.Undo();
+		break;
+	case ME_WINDOW_TREE:
+	case ME_WINDOW_PROP:
+	case ME_WINDOW_PREVIEW:
+	case ME_WINDOW_PREVIEW_PROP:
+	case ME_WINDOW_STAGE:
+	case ME_WINDOW_CONSOLE:
+		materialDocManager.Undo();
+		break;
+	default:
+		break;
+	}
 }
 
 /**
@@ -777,11 +815,21 @@ void MEMainFrame::OnEditUndo() {
 */
 void MEMainFrame::OnEditRedo() {
 	//Check for redo operation on special windows
-	/*if ( ImGui::GetID( "Text" ) == ImGui::GetActiveID() ) {
+	switch ( activeWindow ) {
+	case ME_WINDOW_TEXT_EDIT:
 		m_materialEditView->m_textView.Redo();
-		return;
-	}*/
-	//materialDocManager.Redo();
+		break;
+	case ME_WINDOW_TREE:
+	case ME_WINDOW_PROP:
+	case ME_WINDOW_PREVIEW:
+	case ME_WINDOW_PREVIEW_PROP:
+	case ME_WINDOW_STAGE:
+	case ME_WINDOW_CONSOLE:
+		materialDocManager.Redo();
+		break;
+	default:
+		break;
+	}
 }
 
 /**
--- a/neo/tools/imgui/materialeditor/MEMainFrame.h
+++ b/neo/tools/imgui/materialeditor/MEMainFrame.h
@@ -63,6 +63,8 @@ public:
 	int							OnCreate();
 	void						OnDestroy();
 
+	void						SetActiveWindow(MaterialEditorWindow_t window);
+
 	void						Draw();
 
 	//Menu Message Handlers
@@ -156,6 +158,7 @@ private:
 
 	bool						includeFileInMaterialList;
 
+	MaterialEditorWindow_t		activeWindow;
 };
 
 }
--- a/neo/tools/imgui/materialeditor/MaterialEditView.cpp
+++ b/neo/tools/imgui/materialeditor/MaterialEditView.cpp
@@ -27,6 +27,7 @@ If you have questions concerning this license or the applicable additional terms
 */
 
 #include "MaterialEditView.h"
+#include "tools/imgui/materialeditor/MaterialEditor.h"
 
 #define EDIT_HEIGHT 25
 
@@ -101,16 +102,32 @@ bool MaterialEditView::CanCopy() {
 	return m_textView.CanCopy();
 }
 
+void MaterialEditView::OnCopy() {
+	m_textView.Copy();
+}
+
 bool MaterialEditView::CanCut() {
 	return m_textView.CanCut();
+}
+
+void MaterialEditView::OnCut() {
+	m_textView.Cut();
 }
 
 bool MaterialEditView::CanPaste() {
 	return m_textView.CanCopy();
 }
 
+void MaterialEditView::OnPaste() {
+	m_textView.Paste();
+}
+
 bool MaterialEditView::CanDelete() {
 	return !m_textView.GetReadOnly();
+}
+
+void MaterialEditView::OnDelete() {
+	m_textView.Delete();
 }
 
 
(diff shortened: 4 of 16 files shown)


DanielGibson:
Thank you very much!

I'll look at the code later; are the questions from https://github.com/dhewm/dhewm3/pull/647#issuecomment-2626646805 regarding modal dialogs and layout still open or have you figured that out already?
Modals should be pretty easy (`ImGui::BeginPopupModal()`), for layout I'd have to figure that out myself, depending on what you want to do. Is it about having basically three columns and have that scale when resizing the window?

ashalkhakov:
> I'll look at the code later; are the questions from [#647 (comment)](https://github.com/dhewm/dhewm3/pull/647#issuecomment-2626646805) regarding modal dialogs and layout still open or have you figured that out already? Modals should be pretty easy (`ImGui::BeginPopupModal()`), for layout I'd have to figure that out myself, depending on what you want to do. Is it about having basically three columns and have that scale when resizing the window?

I think I figured modals out. The code:

````
	CDialogPDAAdd dlg;
	if ( dlg.DoModal() == IDOK ) {
/* modal dialog accepted */
	}
````

translates to the opening of the dialog:

````
	addPDADlg.Reset();
	ImGui::OpenPopup( "PDAAdd" );
````

and it's drawing & output handling routine:

````
		if ( ImGui::BeginPopupModal( "PDAAdd", nullptr, ImGuiWindowFlags_AlwaysAutoResize ) ) {
			bool accepted = addPDADlg.Draw();

			if ( accepted ) {
/* modal dialog accepted */
			}

			ImGui::EndPopup();
		}
````

> for layout I'd have to figure that out myself, depending on what you want to do. Is it about having basically three columns and have that scale when resizing the window?

Yes, I wanted to have a 3-column layout like in the original MFC dialog and have it scale or change according to the available space. For this dialog, I think the fixed layout will suffice, for other dialogs we may have to get creative.

There is also the issue that the window position is reset when PDA is selected.

**Comment by ashalkhakov** [comment]
> Ok, I'll have a look!
> 
> By the way, did you check out the ImGui Demo (open Dhewm3Settings menu with F10 -> Other Options -> [Show ImGui Demo])? Maybe it could somehow be done with groups (Layout & Scrolling -> Groups) and/or Tables (Tables & Columns -> Resizable, stretch)?

Thank you, I'll check it out!

**Comment by ashalkhakov** [comment]
@DanielGibson Table layout works nicely! Also, fixed the reappearing window bug: the name of the window passed to ImGui must be stable.

image

**Comment by DanielGibson** [comment]
That's awesome!

Regarding the stable title: You can use "Whatever changing text###PDA Editor" as title - then only "PDA Editor" is used as identifier and only "Whatever changing text" is displayed

File: neo/framework/Common.cpp [hunk 1]
```diff
@@ -1291,7 +1313,7 @@ Com_EditPDAs_f
 ==================
 */
 static void Com_EditPDAs_f( const idCmdArgs &args ) {
```
DanielGibson, line 1315: This should not be hidden behind ID_ALLOW_TOOLS (which is specific to the MFC-tools) anymore, similar to my changes for the light editor.
I think it makes most sense to just always provide that command, even if in (the rare) case that ImGui is disabled it only prints a message that it's not available.

similarly below for `cmdSystem->ADdCommand( "editPDAs", ...)`

DanielGibson: keep in mind that the `#ifndef IMGUI_DISABLE` is needed in here so it still works when not building with ImGui, which especially is the case when building dhewm3ded (the dedicated server)

DanielGibson:
I know this was already the case in  the original PDA editor, but I wonder why the buttons in the left pane use "Add" and "Del" while the others use "Add..." and "Delete..."

Also, I just tried that "Add" (PDA) button and it asked me for a name and I entered it and clicked ok, but the modal didn't close (neither when clicking cancel or pressing escape or anything else). The .pda file was written, though.

DanielGibson:
In https://github.com/DanielGibson/dhewm3/tree/imgui-tools I have (based on your branch) integrated the AfEditor and also modified `Com_EditPDAs_f()` in the way I suggested above

neo/tools/imgui/pdaeditor/PDAEditor.cpp
```diff
@@ -0,0 +1,720 @@
...
+}
+
+void PDAEditor::OnBtnClickedPDAAdd()
+{
+	addPDADlg.Reset();
+	ImGui::OpenPopup( "PDAAdd" );
+}
+
+void PDAEditor::OnBtnClickedPDADel()
+{
```
DanielGibson, line 498: maybe this could be implemented like
```c++
idStr fileName = "newpdas/";
fileName += pdaName; // TODO
fileName += ".pda";
fileSystem->RemoveFile( fileName );
```

ashalkhakov:
> I know this was already the case in the original PDA editor, but I wonder why the buttons in the left pane use "Add" and "Del" while the others use "Add..." and "Delete..."
> 

I just tried to implement deletion, cannot get it to work. The Decl manager keeps a copy of PDA decl in memory, so even if the file is deleted, it will still get re-created.

> Also, I just tried that "Add" (PDA) button and it asked me for a name and I entered it and clicked ok, but the modal didn't close (neither when clicking cancel or pressing escape or anything else). The .pda file was written, though.

Fixed that. I'll add the missing dialogs for audio & video and include this fix.

DanielGibson:
> The Decl manager keeps a copy of PDA decl in memory, so even if the file is deleted, it will still get re-created.

Interesting. Looks like the idDeclManager doesn't support removing decls at all - that's weird.
I guess the Delete button has to remain a dummy then, for now (or maybe even comment it out, what's the point in buttons that don't do anything...)

ashalkhakov:
@DanielGibson Should I do a PR to your branch or to the main repo? (I've just updated `imgui-tools` on my fork.)

DanielGibson:
You could update your tools-porting branch with my changes (either merge or rebase, I don't care) and continue to work on that. I don't think we need a new PR for them, let's just keep this here

neo/tools/imgui/pdaeditor/PDAEditor.cpp
```diff
@@ -0,0 +1,720 @@
...
+
+#include "framework/FileSystem.h"
+#include "framework/Game.h"
+#include "framework/DeclPDA.h"
+
+#ifdef ID_DEBUG_MEMORY
+#undef new
+#undef DEBUG_NEW
+#define DEBUG_NEW new
+#endif
```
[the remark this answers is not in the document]
ashalkhakov: Removed.

DanielGibson:
Great to see you working on the particle editor!

I've toyed a bit with the AF Editor, and it seems to be in a pretty rough shape.
Using it triggers assertions both in dhewm3 code and in ImGui, some things (like the "New Articulated Figure" dialog) are kinda glitchy - not even sure if that thing is "complete" (as in at least *tries* to implement all features).

As I didn't find any bugreports in RBDoom3BFG about that, I guess no one has ever really used it?

Also, not sure if it's easier/faster to fix that thing or to rewrite it (based on the MFC code).

DanielGibson:
I just updated my imgui-tools branch again. It now contains your latest changes and also some new changes from me - most importantly, now if an ImGui window is open, input events (like keys pressed or mouse moved or clicked) are *not* sent to the game but only to ImGui, UNLESS you click click and hold the right mouse button somewhere outside an ImGui window (but in the game window, of course).
This means if for example the light editor is open you can look and move around the level without closing (or even just unfocusing) the light editor window, as long as you hold the right mouse button. And once you stop holding it, the ImGui cursor is back and you can click in the light editor (or other ImGui window) again.

neo/tools/imgui/particleeditor/ParticleEditor.cpp
```diff
@@ -0,0 +1,1585 @@
...
+	listStagesItemData.Add( index, newIndex );
+	listStagesSel = index;
+	EnableStageControls();
+}
+
+idParticleStage *ParticleEditor::GetCurStage() {
+	idDeclParticle *idp = GetCurParticle();
+	int sel = listStagesSel;
+	int index = listStagesItemData.First( sel );
+	if ( idp == NULL || sel == LB_ERR || index >= idp->stages.Num() ) {
```
DanielGibson, line 1188: LB_ERR doesn't make sense here, it's something Windows-specific (and makes the build fail on linux and other operating systems)

neo/tools/imgui/particleeditor/ParticleEditor.cpp
```diff
@@ -0,0 +1,1585 @@
...
+	return instance;
+}
+
+ParticleEditor::ParticleEditor()
+	: colorDlg( "Color" )
+	, fadeColorDlg( "Fade Color" )
+	, entityColorDlg( "Entity Color" )
+	, fileSelection(0)
+	, materialDeclSelection(0)
+	, shouldPopulate(false)
```
DanielGibson, line 94: please make sure that class members are initialized in the same order they appear in the class, at least GCC and clang warn about that kind of thing:
```
In file included from neo/tools/imgui/particleeditor/ParticleEditor.cpp:31:
neo/tools/imgui/particleeditor/ParticleEditor.h: In constructor ‘ImGuiTools::ParticleEditor::ParticleEditor()’:
neo/tools/imgui/particleeditor/ParticleEditor.h:153:41: warning: ‘ImGuiTools::ParticleEditor::entityColorDlg’ will be initialized after [-Wreorder]
  153 |         ColorPicker                     entityColorDlg;
      |                                         ^~~~~~~~~~~~~~
neo/tools/imgui/particleeditor/ParticleEditor.h:143:49: warning:   ‘int ImGuiTools::ParticleEditor::fileSelection’ [-Wreorder]
  143 |         int                                     fileSelection;
      |                                                 ^~~~~~~~~~~~~
neo/tools/imgui/particleeditor/ParticleEditor.cpp:90:1: warning:   when initialized here [-Wreorder]
   90 | ParticleEditor::ParticleEditor()
      | ^~~~~~~~~~~~~~

```

ashalkhakov:
Thanks for support and the suggestions, I will fix them!

As for the progress so far. Regarding dialogs: with classical GUIs it's just a call a dialog as a subroutine, but with ImGui you have to add some extra variables to the state of your app to keep track of which "control state" you're in, i.e. waiting for input from the user or acting on the inputs provided. Not a big deal, it's probably best for the users when there's less modal stuff going on.

More concretely, I'm looking for an autosuggest widget (listbox where you can either select or type the chosen option). I've ported the file dialog code from AF Editor and it works fine.

I've tested the Particle Editor a bit, and it seems to work fine (but need to set `r_useSoftParticles 0` -- we break it when we modify stages in the editor). A few things I want to add:

* sorting of files in the file picker & material picker
* autosuggestion for the file picker & material picker
* sorting of particle systems in the combo box
* use tabs for Particle Editor (I've reproduced the original layout, but the one in The Dark Mod makes more sense and will help cut down the size of the editor window)
* find why requesting all decls is so costly when we populate the particle system combo box. I think it's probably all the parsing that's done. perhaps we can fix this? we could add a method to retrieve just the "header" info about decls

The next steps I'm planning are:

* Decl Browser (simple enough)
* Script editor
* Audio Shader editor (probably as complex as the Particle Editor)

DanielGibson:
> More concretely, I'm looking for an autosuggest widget (listbox where you can either select or type the chosen option).

I think here you want to use an `ImGuiTextFilter` and some kind of ListBox or Table (table makes alternating background colors for the rows easier, if you want that) with Selectables as items.
See `ExampleAppPropertyEditor` in imgui_demo.cpp and also Widgets -> Text Filter and also the "listbox 1" ImGui::ListBox example  under Widgets -> ListBoxes that shows how to use it with ImGui::Selectable

neo/tools/imgui/particleeditor/ParticleEditor.cpp
ashalkhakov: Thanks for noticing. I've fixed it.

neo/tools/imgui/particleeditor/ParticleEditor.cpp
ashalkhakov: Good catch. Fixed.

neo/framework/Common.cpp
```diff
@@ -2331,17 +2384,17 @@ void idCommonLocal::InitCommands( void ) {
...
-	cmdSystem->AddCommand( "editor", Com_Editor_f, CMD_FL_TOOL, "launches the level editor Radiant" );
 	cmdSystem->AddCommand( "editLights", Com_EditLights_f, CMD_FL_TOOL, "launches the in-game Light Editor" );
+	cmdSystem->AddCommand( "editPDAs", Com_EditPDAs_f, CMD_FL_TOOL, "launches the in-game PDA Editor" );
+	cmdSystem->AddCommand( "editAFs", Com_EditAFs_f, CMD_FL_TOOL, "launches the in-game Articulated Figure Editor" );
+#ifdef ID_ALLOW_TOOLS
+	cmdSystem->AddCommand( "editor", Com_Editor_f, CMD_FL_TOOL, "launches the level editor Radiant" );
 	cmdSystem->AddCommand( "editSounds", Com_EditSounds_f, CMD_FL_TOOL, "launches the in-game Sound Editor" );
 	cmdSystem->AddCommand( "editDecls", Com_EditDecls_f, CMD_FL_TOOL, "launches the in-game Declaration Editor" );
-	cmdSystem->AddCommand( "editAFs", Com_EditAFs_f, CMD_FL_TOOL, "launches the in-game Articulated Figure Editor" );
 	cmdSystem->AddCommand( "editParticles", Com_EditParticles_f, CMD_FL_TOOL, "launches the in-game Particle Editor" );
```
DanielGibson, line 2395: remember to also move the lines for editParticles and editScripts above the `#ifdef ID_ALLOW_TOOLS` :)

DanielGibson:
Great work so far!

> but need to set r_useSoftParticles 0 -- we break it when we modify stages in the editor

I couldn't reproduce this, what particle did you use when you had issues with r_useSoftParticles 1?

> find why requesting all decls is so costly when we populate the particle system combo box. I think it's probably all the parsing that's done. perhaps we can fix this? we could add a method to retrieve just the "header" info about decls

You could try setting the `forceParse` argument of `declManager->DeclByIndex()` to `false`.

By the way: I think it would be nice if those lists of materials etc were sorted alphabetically. I think it should work to just call `list.Sort();` - there's an overload for `idList<idStr>` aka `idStrList` that implements a string sort.

And I think it'd be cool if the material picker could show a preview image for a selected material.
This code code from the light editor may come in handy:
```c++
static idImage* GetLightEditorImage( const idMaterial* mat )
{
	// this is similar to what the original light editor does; rbd3bfg iterates through all stages to find a non-null image
	idImage* ret = (mat->GetNumStages() > 0) ? mat->GetStage(0)->texture.image : NULL;
	return ret ? ret : mat->GetEditorImage();
}

void DrawTexture() {
	if( currentTexture != nullptr )
	{
		ImVec2 size( currentTexture->uploadWidth, currentTexture->uploadHeight );

		ImGui::Image( currentTexture->texnum, size, ImVec2( 0, 0 ), ImVec2( 1, 1 ),
			          ImColor( 255, 255, 255, 255 ), ImColor( 255, 255, 255, 128 ) );
	}
}
```

(also, at some point check if `currentTexture->texnum != idImage::TEXTURE_NOT_LOADED` - the light editor just doesn't list any materials without a texnum - and it doesn't only use `mat->GetEditorImage()` but alternatively `mat->GetStage(0)->texture.image` if it exists)

neo/tools/imgui/scripteditor/ScriptEditor.h
```diff
@@ -0,0 +1,107 @@
...
+	//virtual BOOL		PreTranslateMessage( MSG* pMsg );
+	//}}AFX_VIRTUAL
+
+protected:
+	void		OnEditGoToLine();
+	void		OnEditFind();
+	void		OnEditFindNext();
+	void		OnEditReplace();
+	LRESULT		OnFindDialogMessage( WPARAM wParam, LPARAM lParam );
+	void		OnEnChangeEdit( NMHDR *pNMHDR, LRESULT *pResult );
```
DanielGibson, line 71: probably to be expected at this early point of the script editor, but remember that all those FULLCAPS types are windows-only, so this currently breaks the build on Linux and macOS

ashalkhakov: Sorry about that, I'm currently trying to just get something out the door. ^_^ Thanks for the feedback, much appreciated!

ashalkhakov:
> > but need to set r_useSoftParticles 0 -- we break it when we modify stages in the editor
> 
> I couldn't reproduce this, what particle did you use when you had issues with r_useSoftParticles 1?
> 
> You could try setting the `forceParse` argument of `declManager->DeclByIndex()` to `false`.
> 

Thanks! This works great!

> By the way: I think it would be nice if those lists of materials etc were sorted alphabetically. I think it should work to just call `list.Sort();` - there's an overload for `idList<idStr>` aka `idStrList` that implements a string sort.
> 

Should be working now, also, on-the-fly filtering should be working too. ^_^

> And I think it'd be cool if the material picker could show a preview image for a selected material.

Noted! I'll get back to this before finishing up the PR.

ashalkhakov:
@DanielGibson I have a question two questions.

## ScriptEditor

I'm choosing between the following editors to embed:

  * https://github.com/BalazsJako/ImGuiColorTextEdit (or rather, https://github.com/goossens/ImGuiColorTextEdit since it seems more up-to-date)
  * https://github.com/Rezonality/zep/

  You mentioned ImGuiColorTextEdit earlier, but it's not been updated in a while. There are also some tiny Emacs clones floating around that we could embed:

  * [OpenEmacs](https://github.com/DavidSatimeWallin/openemacs)
  * [AttoEmacs](https://github.com/hughbarney/atto)
  * [Med](https://github.com/DigitalMars/med)

  To name a few. We could probably just 're-target' one of these to use ImGui output instead of CLI output since they are so _tiny_.

  Anyways, just looking for feedback. Could you get me up to speed on which one to embed, and what's the process of including it into the codebase? Could I just copy it over? Should I use CMake to pull it from its source repo? Etc.

## Memory management in `PathTreeCtrl`

I'm using `idBlockAlloc` for creating tree nodes, IIUC it doesn't call constructors/destructors for reused nodes? Should I worry about this? Doom's own codebase doesn't seem to use constructors much.

DanielGibson:
Regarding editors: zep also looks good!
There may even be other ImGui textediting widgets with syntax highlighting; just use the one you think works best and is easiest to integrate and has the least (ideally no) additional dependencies.

> We could probably just 're-target' one of these to use ImGui output instead of CLI output

I guess it still would be non-trivial to get ImGui to render this, which includes cursor and text selection, properly (and with good performance).

Not sure of those emacs clones offer that, but it's important that the editor can be used without being familiar with emacs (or vi) shortcuts. Having such a mode as an option is fine, but by default it should behave like a standard GUI text editor (think notepad++, geany, kate, ...).

>  what's the process of including it into the codebase? Could I just copy it over?

Yes, just copy it over, like I do with ImGui.

Regarding the memory management, I'll have to take a closer look, will reply later.

DanielGibson:
> I'm using idBlockAlloc for creating tree nodes, IIUC it doesn't call constructors/destructors for reused nodes? Should I worry about this?

This depends on what types you allocate with it.
If it's just POD (plain-old-data) types that don't need a destructor (like `idVec3`), for example to free dynamically allocated memory (**and** that don't have any members that need a destructor!), it's fine. 
Using it for types that have for example `idList<Foo>` or `idStr` members is bad, because they allocate memory on the heap that gets freed in their destructors, which have to be called.

If your types need idStr or similar, just use regular `new` and `delete` for now.
If it turns out that the default allocators are too slow, we can still try to improve idBlockAlloc or write something else that reuses memory blocks.

*(I first thought that hacking `idBlockAlloc` to call the constructors and destructors would be easy enough, by providing `T* AllocNew()` and `FreeDestruct( T* obj )` functions that do that. But I think `idBlockAlloc` is also expected to free all memory in its own destructor, including objects that had been allocated but not explicitly freed yet - that's harder.)*

neo/framework/Common.cpp
DanielGibson: .. and `Com_EditScripts_f()` and `Com_EditDecls_f()` as well ;)

DanielGibson:
Correction: `idBlockAlloc` *does* call the proper constructors and destructors after all, but not in Alloc() and Free(), but when fresh a `block_t` (that contains an array of `element_t` that has a `type` member) is allocated, or in `Shutdown()` where everything is deleted.

Still not quite the behavior one would (probably) expect..

DanielGibson:
This looks awesome!

One thing that just came to my mind: It would be cool if the text editing widget could also be used for the Script Debugger.
There it should be read-only (only used to render the text with syntax highlighting), but should still allow (right?-)clicking on line numbers to set breakpoints or to tell the debugger to "go on executing until this line" (basically temporary breakpoint).

So if any of the text editing widgets supports "read-only mode that still allows figuring out the line one just clicked in and render something in front of the line number to signify a breakpoint", that one would be preferable :)

ashalkhakov:
@DanielGibson My thoughts exactly! In fact it's the ability to debug things what's drawn me initially to dhewm3. I then found a whole host of other great things, like fully-maintained source code for mods that I can build on. I'll keep an eye on those features.

DanielGibson:
The ImGuiColorTextEdit code needed some small changes to make it build with GCC (and C++11 instead of C++14): [texteditor.patch.txt](https://github.com/user-attachments/files/18762003/texteditor.patch.txt)

(Haven't actually tested the code yet, just made sure it builds because the testbuilds are failing)

ashalkhakov:
> The ImGuiColorTextEdit code needed some small changes to make it build with GCC (and C++11 instead of C++14): [texteditor.patch.txt](https://github.com/user-attachments/files/18762003/texteditor.patch.txt)
> 
> (Haven't actually tested the code yet, just made sure it builds because the testbuilds are failing)

Thanks for the fix! That's something I didn't tell you about. Looks like we'll have to either figure out how to embed Zep (or some other similarly-featured editor), or we'll have to add features to our own fork of ImGuiColorTextEdit. Those forks that I found have the following shortcomings:

* the [original fork](https://github.com/BalazsJako/ImGuiColorTextEdit) has not been updated for a while and there's a scrolling bug in it out of the box (although it still works! and there's a PR fixing that bug)
* the fork by [santaclose](https://github.com/santaclose/ImGuiColorTextEdit) uses C++14 (as you found) and has a dependency on `boost::regex` (we don't need regexes, we already have the code that would be simple enough to port I think) -- I had to remove regex support in the vendored file. it also doesn't support breakpoints and most of the cursor stuff is not public anymore, e.g. one cannot get the selected text.
* the fork by [goosens](https://github.com/goossens/ImGuiColorTextEdit) uses C++17 and I couldn't figure out how to back port it to C++11 (BTW, is this the standard we're using in the codebase?)

None of these have support for:
* go to line (need to implement a custom dialog)
* find/replace (need to implement a custom dialog)
* support for object member completion (custom reaction to keystrokes)
* support for function parameter tooltips (custom reaction to keystrokes)

I've tried to use the single-file inclusion of Zep but it uses C++17 so it doesn't compile, that's a bummer. So I'll probably just go forth and implement those features in our own fork, shouldn't be too difficult, since we already have it implemented in MFC.

DanielGibson:
Regarding the forks: I definitely do *not* want a dependency on boost. Otherwise, use whichever fork works best :)

Regarding the C++ standard: Most of dhewm3 still uses C++98, and unless there's a good reason I'd like to keep it that way so people can build dhewm3 on obscure old platforms like PowerPC Macs.
The ImGui tools currently require C++11, because ImGui itself does. This shouldn't affect old platforms too much, as ImGui can be disabled for them.
If you need C++14 or C++17 for ImGui-tool-specific code, that'd also be fine with me (in my patch I made it build with C++11 instead of configuring C++14 in CMakeLists.txt because it was just one tiny change).
You can bump the C++ standard version in CMakeLists.txt line 236 (`set (CMAKE_CXX_STANDARD 11)`).

DanielGibson:
I've been tweaking the ParticleEditor a bit, so it's narrower, more like the MFC version.
One little thing I noticed was that you use InputFloat() together with SliderFloat(), like this:
image
That's not necessary, or actually redundant - you can just Ctrl+Click into the slider to input the value as text :)
image

Together with my other tweaks it now looks like this:
image

ashalkhakov:
> Regarding the C++ standard: Most of dhewm3 still uses C++98, and unless there's a good reason I'd like to keep it that way so people can build dhewm3 on obscure old platforms like PowerPC Macs.

Okay, that's reasonable and makes my job easier. :)

> The patch implementing these changes: [parteditor.diff.txt](https://github.com/user-attachments/files/18777812/parteditor.diff.txt)

I'll probably add those tabs some time later, in the meantime, this looks much better! So I'll apply it. Thanks!

ashalkhakov:
@DanielGibson Please take a look at the new script editor! There are couple things about it:

* there are no keywords .def files anywhere in the GOG version of Doom 3, so I couldn't test keyword autocompletion
* ESC key doesn't get sent to the text editor, so dismissing autocomplete widget doesn't work

Aside from these two things, it's on feature-parity with the MFC editor aside from multiple windows. The old editor would create multiple OS windows dynamically, but for the new one would it probably makes more sense use a tab control. If you know of something that could work, please let me know.

Next I'd like to tackle either the material or the GUI editor, whichever is simpler.

DanielGibson:
Wow, that's great progress! :)

First, the code requires the following little patch to compile: [imtooltextstuff.patch.txt](https://github.com/user-attachments/files/18900339/imtooltextstuff.patch.txt)

> there are no keywords .def files anywhere in the GOG version of Doom 3, so I couldn't test keyword autocompletion

I don't have those files either, no idea if they ever existed.

> ESC key doesn't get sent to the text editor, so dismissing autocomplete widget doesn't work

That's weird. I'm even using escape in the settings menu.
Does `ImGui::IsKeyPressed( ImGuiKey_Escape, false );` not work? (That's what I use in the settings menu)

> but for the new one would it probably makes more sense use a tab control. If you know of something that could work, please let me know.

The ImGui docking branch (that we're using) supports dragging one imgui window over the other to combine them, which automatically adds a tab bar.
Maybe it's somehow possible to automatically put a window in such a tab?
Compared to standard ImGui tabs that would have the advantage, that you can also drag the a open file *out* of the tab list again, so you can have to scripts (or whatever) open in two ImGui windows, side by side

> Next I'd like to tackle either the material or the GUI editor, whichever is simpler.

Material editor should be easier.

----

I briefly tried the script editor, looks like what's probably supposed to be the autocomplete suggestion box is empty or translucent?  
image

Also, would be nice if double-clicking an entry in the Decl Browser opened it in the editor.

Furthermore, I noticed that editing text is still buggy, especially when using the tab key.
If you put the cursor of the beginning of a line of code and press tab a few times and afterwards press Backspace a few times, it starts deleting characters in the text instead of removing the tabs.

ashalkhakov:
> First, the code requires the following little patch to compile: [imtooltextstuff.patch.txt](https://github.com/user-attachments/files/18900339/imtooltextstuff.patch.txt)

Thanks a ton!

> Does ImGui::IsKeyPressed( ImGuiKey_Escape, false ); not work? (That's what I use in the settings menu)

Just tried it -- it never gets sent to the widget either. :(

> I briefly tried the script editor, looks like what's probably supposed to be the autocomplete suggestion box is empty or translucent?

This should be the keyword suggestion, but since those .def files aren't there we have an empty keyword list. I'll add a small adjustment to only open up the autosuggest box if there are completions.

>  The ImGui docking branch (that we're using) supports dragging one imgui window over the other to combine them, which automatically adds a tab bar. <snip> Compared to standard ImGui tabs that would have the advantage, that you can also drag the a open file out of the tab list again, so you can have to scripts (or whatever) open in two ImGui windows, side by side

Thanks, I'll check this out. I was thinking of exactly this use-case.

> Also, would be nice if double-clicking an entry in the Decl Browser opened it in the editor.

Good point, I'll see to implement this.

> Furthermore, I noticed that editing text is still buggy, especially when using the tab key.

Interesting! Yes, the editor has to go to great lengths due to tab size being 4 spaces. I'll fix this.

FriskTheFallenHuman:
its there any news about this?

ashalkhakov:
Yes I still have the intention to finish this, currently working on the material editor. Just a bit swamped at my job.

ashalkhakov:
I'll commit the latest code today, it contains Material Editor support that is mostly stubbed.

ashalkhakov:
@DanielGibson Could you help with the rendering of `MaterialPreviewView`? How do I render an `idRenderWorld` from inside an ImGui window?

I'm talking about `MaterialPreviewView::Draw`, if the `renderedView.draw` call is uncommented then Dhewm3 crashes in ImGui.

DanielGibson:
Sorry for the late reply, my PC was broken.

Great to see you're still making progress! :)

Side-note: It would be easier for me to investigate this, if the code built on Linux (the automated testbuilds can help finding such problems ;)).

Here are my quick'n'dirty changes to make it build on Linux - at least some of them should be fixed properly (like replace usages of DWORD with an appropriate type). [imgui-tools-linux.diff.txt](https://github.com/user-attachments/files/19528880/imgui-tools-linux.diff.txt)

Regarding your crash (which is actually an assertion, at least it was here): I think the root problem is in `void MEMainFrame::Draw()` line 388 `if ( m_materialPreviewView->Draw( ImVec2( 0, previewSplitterHeight ) ) ) {` - you're passing a width of 0. That makes no sense and probably leads to further problems (many things are calculated with that value)

When fixing that (I just used 200 instead of 0), I get this assertion from ImGui:
`dhewm3: /home/caedes/gamedev/doom3/my-dhewm3/neo/libs/imgui/imgui.cpp:10936: void ImGui::ErrorRecoveryTryToRecoverState(const ImGuiErrorRecoveryState*): Assertion `(0) && "Missing EndChild()"' failed.` 
Unfortunately ImGui only detects this at the end of the frame, so the backtrace doesn't tell where the corresponding BeginChild() was - but this should help narrowing down the problem.

ashalkhakov:
Sorry for the delay. I was thinking of "hijacking" the main window's rendering. If the material editor is the only thing sending commands to the renderer, would that work? The original MFC hides the main window, making it just as inaccessible as this would be.

DanielGibson:
I realized that you'll most probably have to render to a texture and then render that with ImGui.

The reason is that thanks to the ImGui docking branch with multi-viewports, which means that ImGui windows can be dragged out of the Doom3 (SDL/operating-system-level) window which implicitly creates a new SDL window and then ImGui renders in there - but I think it would be hard to make Doom3 render to multiple windows (much harder than rendering to a specific region of the main window). So it's probably most feasible to render to a texture and then use `ImGui::Image()` with `idImage::texNum`.

ashalkhakov:
@DanielGibson I'm thinking of removing the Preview View and instead modifying game code slightly to allow for a custom material on testmodel, similarly to what particle editor is doing. I think it will be way easier to do, and it will allow editing a material in-game. I'm not sure if there are any downsides to this, though.

ashalkhakov:
Actually, it seems like it kind of works as-is. Would you kindly take a look? @DanielGibson 

Here's a screenshot:

image

I've pushed these changes. The preview window shows a portion of the main window though, and it's inverted vertically. Also, the test model should be a box but it's clearly a low poly sphere which shows up when typing `materialEditor` in the console. Please also note that I've used `testparticlematerial` as a material.

* why is it upside down?
* why does it overlap with the main menu's 3D view?
* why is the model incorrectly displayed?

(BTW I've gotten the ASE file from an LLM so it might be wrong, I didn't find the ASE models referenced by the material editor among D3 files)

FriskTheFallenHuman:
> Actually, it seems like it kind of works as-is. Would you kindly take a look? @DanielGibson
> 
> Here's a screenshot:
> image
> 
> I've pushed these changes. The preview window shows a portion of the main window though, and it's inverted vertically. Also, the test model should be a box but it's clearly a low poly sphere which shows up when typing `materialEditor` in the console. Please also note that I've used `testparticlematerial` as a material.
> 
>     * why is it upside down?
> 
>     * why does it overlap with the main menu's 3D view?
> 
>     * why is the model incorrectly displayed?
> 
> 
> (BTW I've gotten the ASE file from an LLM so it might be wrong, I didn't find the ASE models referenced by the material editor among D3 files)

@ashalkhakov not need to use that ASE model, id forgot to ship the ASE models for Material Editor, they where later added for the RoE expansion but never include in base, you could  try reference these.

EDIT: in case you need them here they are: 
[missing_materialeditor_models.zip](https://github.com/user-attachments/files/21804186/missing_materialeditor_models.zip)
Ideally these should be include in the base/ folder but that is up to Daniel

DanielGibson:
> Ideally these should be include in the base/ folder but that is up to Daniel

So far I avoided shipping any gamedata, and probably the license of that is also problematic..

OTOH the next dhewm3 release (and the current git version) supports widescreen-capable menus, and to use that one needs adjusted menu files, so I'll have to ship gamedata that is not "free" (but derived from the commercial Doom3 data) anyway.

However, as those ASE files are not under a GPL-compatible license, they should not be part of this repo.
We could think about creating a "dhewm3-gamedata" repo or similar, or just create a .pk4 file that's available for download with the releases.

@ashalkhakov do those problems persist when using the proper ASE files?

ashalkhakov:
@DanielGibson Yes things work fine with the proper ASE files. We could also hard-code them though? The files contain boxes (of different sizes), cylinders (2 sizes) and a sphere.

DanielGibson:
Yeah, hard-coding those kind of geometries would be a nice solution

FriskTheFallenHuman:
Any news about this?

ashalkhakov:
I’m currently porting the debugger.I will probably progress much faster on Christmas holidays.28 жовт. 2025 р. о 20:33 Krispy ***@***.***> пише:﻿FriskTheFallenHuman left a comment (dhewm/dhewm3#649)
Any news about this?

FriskTheFallenHuman:
are there plans for having a gui editor in imgui?

ashalkhakov:
Yes, I want to have full feature parity with the old editors.

PR #650, opened 2025-02-03
neo/CMakeLists.txt: added use of O3 instead of O2 for e2k arch

r-a-sattarov:
Using O3 option on [E2K](https://en.wikipedia.org/wiki/Elbrus_(computer)) mcst-lcc compiler approximately equal to O2 at X86/ARM gcc.

### DIFF
--- a/neo/CMakeLists.txt
+++ b/neo/CMakeLists.txt
@@ -329,11 +329,18 @@ if(D3_COMPILER_IS_GCC_OR_CLANG)
 		endif ()
 	endif ()
 
+	if(cpu STREQUAL "e2k" AND CMAKE_COMPILER_IS_GNUCC)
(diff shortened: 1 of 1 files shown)


EntityFX:
Looks good

DanielGibson:
if you say so.. it's not like *I* have to debug that shit if weird issues happen due to aggressive optimization (because I don't have such a CPU) ¯\\_(ツ)_/¯

r-a-sattarov:
The O3 optimization level on MCST-LCC is almost equal to O2 on GCC. So everything will be fine.
And O4 on MCST-LCC (yes, there is one) is approximately equal to O3 GCC.

Ref: http://mcst.ru/doc/elbrus_prog/html/chapter1.html#lcc

Issue #651, opened 2025-02-03
Low Fps with map UAC Franchise 666


walikor:
hi, I was trying some doom 3 maps like zombie_shuffle (works perfectly) and UAC Franchise 666, which unfortunately as soon as map 6661 starts the frame rate drops to 4 fps and gives errors on the console.
This is the map: https://www.moddb.com/games/doom-iii/addons/uac-franchise-666
and this is the image.
with vanilla doom 3 work well.
![Image](https://github.com/user-attachments/assets/7f0c8445-ed51-45e6-bce6-5911d1331103)

DanielGibson:
> with vanilla doom 3 work well.

does this mean that this warning isn't printed in vanilla doom3, or just that it runs at a higher framerate?

For a higher framerate it often helps to disable soft particles and reduce anti-aliasing (esp. 16x is very demanding even on todays hardware)

Also: what operating system are you using, what hardware, what version of dhewm3?
All that text that is in the issue report form when creating a new issue is there for a reason...

walikor:
> > with vanilla doom 3 work well.
> 
> does this mean that this warning isn't printed in vanilla doom3, or just that it runs at a higher framerate?
> 
> For a higher framerate it often helps to disable soft particles and reduce anti-aliasing (esp. 16x is very demanding even on todays hardware)
> 
> Also: what operating system are you using, what hardware, what version of dhewm3? All that text that is in the issue report form when creating a new issue is there for a reason...

exactly with doom 3 vanilla I don't have those errors in the console (no valid point found) and the framerate remains at 62, I disabled the soft particles and set the anisotropic filter to 1x but it didn't work, I'm using:
-windows 10 22H2
-Intel i5 8500 (6 cores 6 threads)
-32GB Ram DDR4
-NVIDIA GeForce RTX 4070 SUPER
-Dhewm3 Version 1.5.4

DanielGibson:
I can reproduce this now (only on Windows, for whatever reason), I'm investigating...

DanielGibson:
See https://github.com/dhewm/dhewm3/pull/652 for a fix, please test :)

walikor:
> See [#652](https://github.com/dhewm/dhewm3/pull/652) for a fix, please test :)

hi, I tested with both the 32 bit and 64 version and they work correctly, without drops in fps or errors in the console.
Thanks.

DanielGibson:
thanks for testing, I merged the fix

DanielGibson:
[...]

PR #652, opened 2025-02-04
Fix issues in first map of the "UAC Franchise 666" user campaign, #651

DanielGibson:
somehow the collision code managed to spread NaNs on Win32, which caused a horrible framerate, "GetPointOutsideObstacles: no valid point found" warnings in the console and assertions in debug builds. Didn't happen in Vanilla Doom3 though.

At the location I changed the code in, I saw the following values in the debugger:
normal: {x=0.00610326231 y=5.58793545e-09 z=1.19209290e-07 } trmEdge->start: {x=-1358.00000 y=913.948975 z=25.2637405 } start: {x=-1358.00000 y=916.000000 z=34.0000000 }
end:   {x=-1358.00000 y=810.000000 z=34.0000000 }
dist (normal*trmEdge->start): -8.28822231
d1: 9.53674316e-07
d2: 9.53674316e-07
f1 (d1/(d1-d2)): inf

"normal" isn't normalized and also very small (in all directions), "start" and "end" have quite different y values, but still doing scalar multiplications of each with "normal" gave the same result.. No idea what this all means exactly, but checking if d1 - d2 is (almost) 0 to prevent INF solved the problems. In the end it will be some tiny differences in floating point calculations between different platforms and compilers..
In my test d1-d2 was exactly 0, but I compare with FLT_EPSILON to be on the safer side.

### DIFF
--- a/neo/cm/CollisionModel_translate.cpp
+++ b/neo/cm/CollisionModel_translate.cpp
@@ -319,7 +319,10 @@ void idCollisionModelManagerLocal::TranslateTrmEdgeThroughPolygon( cm_traceWork_
 			dist = normal * trmEdge->start;
 			d1 = normal * start - dist;
 			d2 = normal * end - dist;
-			f1 = d1 / ( d1 - d2 );
+			float d1d2diff = d1 - d2;
+			// DG: d1 - d2 was 0 in some weird case, which caused f1 to be INF,
+			//     which caused NaN mayhem all over the place
+			f1 = ( fabsf(d1d2diff) > idMath::FLT_EPSILON ) ? d1 / d1d2diff : 0.0f;
 			//assert( f1 >= 0.0f && f1 <= 1.0f );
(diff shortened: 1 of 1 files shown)


DanielGibson:
This should fix the issues from #651.
You can find testbuild for Windows at https://github.com/dhewm/dhewm3/actions/runs/13127361349?pr=652 under "Artifacts" (dhewm3-win32-031b206)

@walikor please test

walikor:
> This should fix the issues from #651. You can find testbuild for Windows at https://github.com/dhewm/dhewm3/actions/runs/13127361349?pr=652 under "Artifacts" (dhewm3-win32-031b206)
> 
> @walikor please test

hi, I tested with both the 32 bit and 64 version and they work correctly, without drops in fps or errors in the console.
Thanks again.

DanielGibson:
Great, thanks for testing!

PR #653, opened 2025-02-17
Improvements in support for cst anchors in GUIs

DanielGibson:
We have the following issues:
- [x] "scaleto43 0" in windowDef should disable scaling to 4:3 (with black/empty bars) for fullscreen menus, so ones with anchors can cover the whole screen
- [x] renderDefs don't properly support anchor stuff (get stretched when they shouldn't)
- [x] mouseclicks go to the wrong coordinates (next to button) when using anchors in non-4:3 resolutions

This PR is a draft until all three issues are fixed

----

Support "scaleto43 0" in Desktop windowDefs to explicitly stretch them

even if r_scaleMenusTo43 1 and they'd usually be scaled to 4:3. Useful when using anchors in fullscreen menus (like the PDA)

This is analogous to "scaleto43 1" in windowDefs which allows scaling to 4:3 (with black/empty bars) even if by default it would *not* be scaled.


DanielGibson:
I used this hacked mainmenu from Fitz Packerton for testing the latest commit (because Doom3s mainmenu is way too complex for debugging): [mainmenu.gui.txt](https://github.com/user-attachments/files/18839070/mainmenu.gui.txt)

DanielGibson:
When using a windowDef anchored to the center, in widescreen resolutions there are empty spaces on the left and right side.
To allow filling those spaces with other windowDefs (that do not overlap with the center one, no matter how wide the empty area is), I added a window variable "gui::horPad" that contains the current width of those areas (in the usual coordinates).
So for example
```
	windowDef leftPadFiller {
		cstAnchor  CST_ANCHOR_LEFT
		rect   0, 0, "gui::horPad", 480
		backColor 0, 1, 0, 1
	}
```
would fill the left area with green and

```
	windowDef leftPadFiller {
		cstAnchor  CST_ANCHOR_RIGHT
		rect   0, 640 - "gui::horPad", "gui::horPad", 480
		backColor 0, 1, 0, 1
	}
```
does the same for the right area.

I also added "gui::vertPad" that does the same for the top/bottom area when using thin/tall resolutions.

Here's a sample gui: [minmain.gui.txt](https://github.com/user-attachments/files/18880274/minmain.gui.txt)

It looks like this (with different resolutions of the dhewm3 window):
shot00144
shot00141
shot00142
shot00143

DanielGibson:
Another cool thing possible with this: An unstretched tiling background that goes over the whole screen:
```
	windowDef unstretchedFullSizeBackground {
		cstAnchor  CST_ANCHOR_TOP_LEFT
		cstNoClipBackground 1 // allows to use more than 640x480
		// this rect spans the whole screen
		rect       0, 0, 640 + 2 * "gui::horPad", 480 + 2 * "gui::vertPad"
		background	"gui/mainmenu/star"
		naturalmatscale 1 // to make the texture tile
		// better don't use mouse cursor events on such extreme rect resolutions:
		noevents   1
	}
```

image
image
image

this is basically minmain.gui from the previous post, but with the change from this one and with leftPadFiller and rightPadFiller commented out so the background is more visible

DanielGibson:
I just renamed "gui::horPad" to "gui::cstHorPad" and "gui::vertPad" to "gui::cstVertPad", so it's clearer that they belong to the CST anchor system.

I also added "gui::cstAspectRatio" and "gui::cstWidth" and "gui::cstHeight", for convenience.

The latter are the width and height of the screen (or Doom3 operating-system-level window),
but not in physical pixels but in those **virtual coordinates* used by the GUIs, based on 640x480.
For widescreen resolutions, "gui::cstHeight" is always 480 and "gui::cstWidth" is scaled according to the aspect ratio (>640).
For tall resolutions (narrower than 4:3), "gui::cstWidth" is always 640 and "gui::cstHeight" is scaled according to the aspect ratio (> 480).

PR #656, opened 2025-02-25
Fix memory deallocation issue by using delete[] for arrays

klaussilveira:
This PR fixes an issue in the `idODE_Midpoint` destructor where dynamically allocated arrays were incorrectly deallocated using `delete` instead of `delete[]`, leading to undefined behavior.


DanielGibson:
great find, thank you! :)

PR #658, opened 2025-02-26
Fix buffer overflow warning when calculating game time.

klaussilveira:


neo/d3xp/MultiplayerGame.cpp
```diff
@@ -861,7 +861,16 @@ const char *idMultiplayerGame::GameTime() {
 		t = s / 10;
 		s -= t * 10;
 
-		sprintf( buff, "%i:%i%i", m, t, s );
+		// Format time as m:ts (minutes:tens-seconds)
```
DanielGibson, line 864: please don't change the behavior of the code

neo/game/MultiplayerGame.cpp
```diff
@@ -554,14 +554,22 @@ const char *idMultiplayerGame::GameTime() {
...
+		m = s / 60;
+		s -= m * 60;
+		t = s / 10;
+		s -= t * 10;
 
-		sprintf( buff, "%i:%i%i", m, t, s );
+		// Format time as m:ts (minutes:tens-seconds)
+		// Ensure we never exceed buffer size by using idStr::snPrintf
+		if ( m >= 100 ) {
+			// If minutes exceed 99, just show 99:59
```
DanielGibson, line 566: why?

neo/d3xp/MultiplayerGame.cpp
DanielGibson: (or does this fix a real issue at >= 100 minutes?)

neo/game/MultiplayerGame.cpp
DanielGibson: Why?
Why not show e.g. 120 minutes for a game that goes on for two hours?
Also, it doesn't go to infinity but only up to 35'791 minutes, as the calculation in the comments that you removed showed - then ms wraps around and becomes negative, but is set to `0` in the code above.
But it does limit the maximum length of the resulting string and guarantees that the buffer won't overflow (and 35791 minutes are 24 days)

DanielGibson:
To be clear, I don't mind replacing sprintf() with snprintf() to shut up the compiler warnings, I'm just against changing the behavior

PR #659, opened 2025-02-26
Removed unused code

klaussilveira:


neo/renderer/Material.cpp
```diff
@@ -2082,10 +2082,9 @@ void idMaterial::ParseMaterial( idLexer &src ) {
 	if ( cullType == CT_TWO_SIDED ) {
 		for ( i = 0 ; i < numStages ; i++ ) {
 			if ( pd->parseStages[i].lighting != SL_AMBIENT || pd->parseStages[i].texture.texgen != TG_EXPLICIT ) {
-				if ( cullType == CT_TWO_SIDED ) {
-					cullType = CT_FRONT_SIDED;
-					shouldCreateBackSides = true;
-				}
+				cullType = CT_FRONT_SIDED;
+				shouldCreateBackSides = true;
```
DanielGibson, line 2086: Could you explain the purpose of this change and why you think it's correct?
It's not immediately obvious to me.

DanielGibson: ah nevermind, now I see, `if ( cullType == CT_TWO_SIDED )` was already checked before the loop

DanielGibson:
thanks! :)

PR #660, opened 2025-02-26
Fix uninitialized variable warning in R_IssueEntityDefCallback

klaussilveira:


neo/renderer/tr_light.cpp
```diff
@@ -1062,9 +1062,12 @@ R_IssueEntityDefCallback
 */
 bool R_IssueEntityDefCallback( idRenderEntityLocal *def ) {
 	bool update;
-	idBounds	oldBounds;
+	idBounds oldBounds;
 	const bool checkBounds = r_checkBounds.GetBool();
 
+	// Initialize oldBounds to avoid the uninitialized variable warning
```
DanielGibson, line 1068: it's not used uninitialized, it's only used `if ( checkBounds )`, and in that case it's initialized

DanielGibson:
Weirdly, on my system neither g++ 13.3 nor clang++ 18 warn about this - even if I *do* make it uninitialized by commenting out `oldBounds = def->referenceBounds;`

PR #666, opened 2025-03-22
Fix building without imgui

coldtobi:
If neo/libs/imgui is removed, the build fails, even when IMGUI is set to "Off". Moving the include inside the IMGUI_DISABLE guard makes the building OK.

Reproducer:

```
 rm -r neo/libs/imgui
 mkdir build
 cd build
 cmake -DIMGUI=Off ../neo/
 make
```

Error:

```
[ 17%] Building CXX object CMakeFiles/dhewm3.dir/framework/Common.cpp.o /usr/lib/ccache/c++ -DD3_ARCH=\"x86_64\" -DD3_HAVE_LIBBACKTRACE -DD3_OSTYPE=\"linux\" -DD3_SIZEOFPTR=8 -DID_REPRODUCIBLE_BUILD -DIMGUI_DISABLE -D__DOOM_DLL__ -I/usr/include/AL -I/usr/include/SDL2 -I/home/tobi/workspace/deb/packages/games-team/dhewm3/dhewm3/obj-x86_64-linux-gnu -I/home/tobi/workspace/deb/packages/games-team/dhewm3/dhewm3/neo -I/home/tobi/workspace/deb/packages/games-team/dhewm3/dhewm3/neo/libs/imgui -g -O2 -ffile-prefix-map=/home/tobi/workspace/deb/packages/games-team/dhewm3/dhewm3=. -fstack-protector-strong -fstack-clash-protection -Wformat -Werror=format-security -fcf-protection -Wdate-time -D_FORTIFY_SOURCE=2 -Woverloaded-virtual -Wno-class-memaccess -Wno-c++20-compat -pipe -Wall -fno-strict-aliasing -ffp-contract=off -fvisibility=hidden -Wno-sign-compare -Wno-switch -MD -MT CMakeFiles/dhewm3.dir/framework/Common.cpp.o -MF CMakeFiles/dhewm3.dir/framework/Common.cpp.o.d -o CMakeFiles/dhewm3.dir/framework/Common.cpp.o -c /home/tobi/workspace/deb/packages/games-team/dhewm3/dhewm3/neo/framework/Common.cpp
In file included from /home/tobi/workspace/deb/packages/games-team/dhewm3/dhewm3/neo/framework/Common.cpp:53: 
/home/tobi/workspace/deb/packages/games-team/dhewm3/dhewm3/neo/sys/sys_imgui.h:6:10: 
fatal error: ../libs/imgui/imgui.h: No such file or directory
    6 | #include "../libs/imgui/imgui.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/neo/sys/sys_imgui.h
+++ b/neo/sys/sys_imgui.h
@@ -3,8 +3,6 @@
 #ifndef NEO_SYS_SYS_IMGUI_H_
 #define NEO_SYS_SYS_IMGUI_H_
 
-#include "../libs/imgui/imgui.h"
-
 namespace D3 {
 namespace ImGuiHooks {
 
@@ -16,6 +14,8 @@ enum D3ImGuiWindow {
 };
 
 #ifndef IMGUI_DISABLE
+
+#include "../libs/imgui/imgui.h"
 
 extern ImGuiContext* imguiCtx; // this is only here so IsImguiEnabled() can use it inline
 


coldtobi:
mmm... strange why it does fail here, and not on my tries.. I'll investigate, (but that will be another day, EBUSY)

DanielGibson:
Why would you remove part of the source?

coldtobi:
to ensure that the embedded code copy is not used. Debian policy.

DanielGibson:
debian policy is to remove features from software?

coldtobi:
dhewm3 works well without imgui.

That said, long term plan is to use the packaged imgui for it, but I'm hesitant to make this change so short before the freeze  (freeze for Debian 13 is imminent.)

DanielGibson:
Packaged imgui is just not feasible.
imgui gets frequent updates and basically every update breaks the API.
Projects using it update to new versions at different speeds.
Trying to make all projects use the same ImGui version, the one packaged by debian, would be quite hard for the maintainers of the debian packages of those projects, and would probably result in broken software as projects rely in features or bugfixes from newer versions.

neo/sys/sys_imgui.h
```diff
@@ -17,6 +15,8 @@ enum D3ImGuiWindow {
 
 #ifndef IMGUI_DISABLE
 
+#include "../libs/imgui/imgui.h"
```
DanielGibson, line 18: you're moving the include into `namespace D3::ImGuiHook` - that can't work, of course.

DanielGibson:
Fixed in https://github.com/dhewm/dhewm3/commit/42cfb157a22a09a22527aecf5ae5bfa47a1a375d

coldtobi:
> Packaged imgui is just not feasible. imgui gets frequent updates and basically every update breaks the API. Projects using it update to new versions at different speeds. Trying to make all projects use the same ImGui version, the one packaged by debian, would be quite hard for the maintainers of the debian packages of those projects, and would probably result in broken software as projects rely in features or bugfixes from newer versions.

Packaging ImGui is feasible and already done by several distros. While ImGui updates frequently, API breakage is usually minor, and projects can update incrementally. Many libraries evolve similarly, yet distributions package them successfully. Coordinating versions is a common task for maintainers, and Debian already handles libraries with varying update cadences. Debian avoids code copies to ensure security fixes and improvements are applied consistently across all packages, reducing maintenance burden and preventing outdated or vulnerable code from lingering in the system. Furthermore, Debian's primary goal is to provide a stable release, not to chase every new upstream update. 
Said that important bug fixes and security fixes are backported to the versions in stable. 
Yes, that means more work for me as maintainer, but Debian's focus is a bit different here.  Experience shows that this works well in the Debian universe.

DanielGibson:
And then there's details like ImGui standard branch vs Docking branch, it using stb_truetype.h vs libfreetype and probably other compiletime differences that users of the lib depend on.
And patches to it, for example dhewm3 has this change: https://github.com/dhewm/dhewm3/blob/42cfb157a22a09a22527aecf5ae5bfa47a1a375d/neo/libs/imgui/backends/imgui_impl_opengl2.cpp#L75-L111
so it uses qgl (the integrated OpenGL function loader) - and this probably is a problem in general, because different projects don't only use different rendering backends (OpenGL1/2, OpenGL3, Vulkan, ...) and platform abstractions (SDL2, SDL3, GLFW, ...) but also different loaders for OpenGL/Vulkan functions.

PR #669, opened 2025-04-08
Allow configuring "Only use precompressed textures if they're BPTC (BC7)"

DanielGibson:
Allow configuring "Only use precompressed textures if they're BPTC (BC7)" by setting `image_usePrecompressedTextures 2`.

Also allow configuring "compress uncompressed textures to BPTC on upload to GPU" by setting `image_useCompression 2`.

And expose that in the Dhewm3SettingsMenu


DanielGibson:
Test binaries for Windows are available at https://github.com/dhewm/dhewm3/actions/runs/14344017719?pr=669
(under Artifacts; you need to be logged in at Github to download them)

Issue #698, opened 2025-10-14
MD3 issues


MattLR:
I've been porting a Q3 based game over to Doom 3 and I've run into a couple of problems with directly using the MD3 models.

The biggest issue is that a missing MD3 model will (or at least can) cause the game and editor to crash on loading with no feedback.

The second issue is that MD3 models don't cast shadows in game, they seem to receive lighting correctly otherwise.

DanielGibson:
Can you post models (incl. textures, .def files and whatever else is required) that help reproducing the issues?

MattLR:
Here are some example maps, from my testing it's just the md3 extension that causes crashes, missing ase/lwo work fine as you would expect, and invalid (eg .whatever) file extensions are also handled properly.

https://drive.google.com/file/d/1qUAaA1Kz8G9fGpiqpiW4WN06DKP42YhV/view?usp=drive_link

Lighting issue appears to be common across all MD3 files from testing.

I also found that the statue MD3 model doesn't have any collision whereas the ASE does, I hadn't noticed this before.

I provided two different maps with missing MD3 models because there's a slight difference in behaviour between vanilla Doom3/fhDoom and dhewm3, dhewm3 will crash even if the MD3 file is in an invalid entity (misc_model), whereas vanilla/fhDoom it seems to only crash when the entity would actually load the model, although I only tested it with func_static.

test_models - lighting/collision problems

duel8_sample_ase/md3 - map copied from JA that will load as long as the models are renamed from .md3

test_missing_models - crash on loading example with func_static instead of misc_model for the slightly different behaviour in vanilla D3

DanielGibson:
Thanks for the testcases!

Regarding the collision (and maybe even lighting?) problems: Have you tried more options in the material?
Especially `solid` and/or `collision` seem worth a try.
(My theory why ASE and MD3 behave differently is that MD3 models are considered dynamic, while ASE are just one kind of model that's loaded into the generic `idRenderModelStatic`)

Regarding the crashes, I have a workaround (fix?) in #700
I wonder if only missing MD3 models cause these crashes, or if using MD5 (in the same way) would also crash or maybe spam the console if the file can't be found

MattLR:
I tested all the material keywords I could think of that could affect shadows and collision, as far as I could tell it didn't affect how the models were treated.

[...]

PR #700, opened 2025-10-22
Fix MD3 issues

DanielGibson:
#698 reports some issues with MD3 files:
- [x] missing MD3 files crash on level load (instead of using a default black box model)
- [x] MD3 models don't cast shadows
- [x] editor crashes with missing MD3 files
- [ ] MD3 models don't have collision

At least the crash should be fixed now

### DIFF
--- a/neo/renderer/Model_md3.cpp
+++ b/neo/renderer/Model_md3.cpp
@@ -321,9 +321,11 @@ idRenderModel *idRenderModelMD3::InstantiateDynamicModel( const struct renderEnt
 	int				frame, oldframe;
 	idRenderModelStatic	*staticModel;
 
+	int numFrames = md3->numFrames;
+
 	// TODO: these need set by an entity
-	frame = ent->shaderParms[SHADERPARM_MD3_FRAME];			// probably want to keep frames < 1000 or so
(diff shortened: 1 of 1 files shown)


DanielGibson:
Testbinaries for Windows can be downloaded at https://github.com/dhewm/dhewm3/actions/runs/18702760746?pr=700 (under "Artifacts")

DanielGibson:
Win Testbuilds with better solution for shadow issue: https://github.com/dhewm/dhewm3/actions/runs/18763962423?pr=700

DanielGibson:
New testbinaries: https://github.com/dhewm/dhewm3/actions/runs/18810805963?pr=700

@MattLR can you test those? If no new issues turn up I'd merge this (even if the selection in editor issue persists I don't feel like investigating it any further, maybe someone more familiar with the tool code can do that..)

MattLR:
I've been using the test binaries for the last couple of days and it's been working great so far.

Just as a note for if the MD3 selection issue gets looked into later on, you actually can select the models now in both the 2D and 3D views, but you have to click directly on the origin in 2D and very close to it in 3D (and it seems to be much harder to select larger models in the 3D view, but that might just be because eyeballing the origin is much harder).

PR #585, opened 2024-06-27
WIP: Support >60Hz (2nd try)

DanielGibson:
Based on @dezo2's [patch](https://github.com/DanielGibson/dhewm3/commit/46e519f24556fd87c87d2133896ce75c6399355d) ([see also](https://github.com/dhewm/dhewm3/issues/250)), but adjusted to make the framerate configurable with the `com_gameHz` CVar + my own fixes, also some ideas from Stradex' branch.

Very much WIP, and no guarantees this will work out

----
## TODO
Potential bugs that have been observed with various attempts of running dhewm3 with >60Hz and that are worth testing:

- [ ] **Port Mods** - they currently don't work because the game API changed. Only the base game and Resurrection of Evil are supported
- [x] [Crashes with crane in Alpha Labs Sector 3](https://github.com/dhewm/dhewm3/issues/230#issuecomment-502461387) (should be fixed by patching GAME_FPS and GAME_FRAMETIME in the scripts)
    - Should *really* be fixed now, was actually something with SetSpeed() in C++ code that I first fixed incorrectly
- [x] Chaingun shooting too fast (should be fixed by patching CHAINGUN_FIRE_SKIPFRAMES in script)
- [x] [Flickering particles](https://github.com/dhewm/dhewm3/issues/261) (happened with `com_fixedTic -1`, might not be relevant here?)
    - Haven't seen this issue so far (and no one else has reported it either)
- [x] [Switching Mods/game modes can cause black menu for multiple seconds to minutes](https://github.com/simonedibilio/dhewm3/issues/2)
    - ~~I can *still* reproduce this bug. I think it's related to counting tics (`com_ticNumber`?) and then multiplying it with USERCMD_MSEC and comparing the result to Sys_Milliseconds() or something, combined with one mod having a different `com_gameHz` value than the other, so tics are longer/shorter. Needs more investigation..~~  
    Should be fixed 
- [x] [RoE gravity gun only works after 2 minutes or so](https://github.com/simonedibilio/dhewm3/issues/2) (haven't tried this yet, might be similar to the previous one)
    - I couldn't reproduce this with my branch, if anyone runs into this problem please report!
- [x] [I found if I tweak the fps higher than the monitor refresh rate, NPCs and mobs in-game will no longer interact with stairs. I think this is not a bug related to this fork, cause I found it in BFG Edition too.](https://github.com/simonedibilio/dhewm3/issues/2) (again from same bugreport on simonedibilio's branch - this sounds extremely weird)
[...]

PR #701, opened 2025-10-27
Smoother timing (reduce micro stutters at 60fps)

DanielGibson:
I ported some code from #585 that increases timer accuracy and introduces `com_showFPS 2` which also shows avg/min/max frametimes of the last 10 frames.

If you've ever noticed microstutters in dhewm3 (esp. when VSync is enabled), please test this branch and see if it improves things!

### DIFF
--- a/docs/GUIs.md
+++ b/docs/GUIs.md
@@ -182,7 +182,7 @@ Luckily, dhewm3 has some improvements :)
 
 dhewm3 has several features to better support non-4:3 resolutions.
 
-### Scaling of mMenus to 4:3 with Letter-Boxing
+### Scaling of Menus to 4:3 with Letter-Boxing
 
 *dhewm3* scales fullscreen menus like the main menu and the PDA to 4:3 regardless of the resolution,
 leaving **black bars on the left and the right**, or if you happen to use a portrait resolution like
@@ -347,7 +347,7 @@ If the screen has 4:3 aspect ratio, anchoring to top-left looks like this:
 ![](img/d3gui-cst-tl-43.png)
 
 Note how:
-* In non 4:3 modes he `Desktop` is stretched (=> its grid isn't square).
+* In non 4:3 modes the `Desktop` is stretched (=> its grid isn't square).
 * The `AnchorArea` retains its 4:3 aspect ratio and fills as much of the screen (or dhewm3 window)
   as possible without stretching, staying at the *top* or *left* border of the screen, like the name 
   `CST_ANCHOR_TOP_LEFT` suggests.


DanielGibson:
**NOTE: These builds are outdated, see in a comment below for new links!**

~~Here are testbuilds for Windows: https://github.com/dhewm/dhewm3/actions/runs/18830049016?pr=701 (under "Artifacts")~~

~~Mirroring the one for Win32 here so you can download it without logging into Github: [dhewm3-win32-5558e76.zip](https://github.com/user-attachments/files/23156843/dhewm3-win32-5558e76.zip)~~

neo/framework/Common.cpp
```diff
@@ -246,12 +246,29 @@ class idCommonLocal : public idCommon {
 	idCompressor *				config_compressor;
 #endif
 
-	SDL_TimerID					async_timer;
+	static int					AsyncThread(void* arg);
+	xthreadInfo					asyncThread;
+	volatile bool				runAsyncThread;
```
turol: `volatile` is not recommended for multithreading. Prefer `std::atomic`.

https://stackoverflow.com/questions/2484980/why-is-volatile-not-considered-useful-in-multithreaded-c-or-c-programming

DanielGibson: That would require C++11, though.
I know this is not exactly clean, but I also don't think it can cause trouble: The only time that bool is changed while the thread is running is when it's set to `false` on shutdown.
Even if the modification is not atomic (however that should happen in a byte-sized bool), it doesn't matter if that thread runs one more frame than it technically should - eventually it will see the changed value (as `volatile` makes sure it's really re-read on each iteration).

neo/framework/Session.cpp
```diff
@@ -533,6 +533,8 @@ void idSessionLocal::CompleteWipe() {
 	}
 }
 
+extern void Com_UpdateFrameTime();
```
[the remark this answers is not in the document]
DanielGibson: Maybe.. it's only used in Common.cpp and this one place so I didn't want to make it too visible... and as this is C++ at least linking will fail if the function signature ever changes

DanielGibson:
I realized that this still isn't optimal and that I should probably get rid of the async thread after all.. haven't finished that yet though, because doing it (or finding the right place to do it) is non-trivial due to different code paths that use `com_ticNumber` and rely on it being updated even if common->Frame() might not be run in the mean time, session->Frame() not being run at all in MP mode (even though it has a lot of the frame timing logic in it) and more..

And on top of that I want it to coordinate with vsync, that's actually the main reason I want to modify the time/place `com_ticNumber` is updated: That update (and triggering `TRIGGER_EVENT_ONE` which `session->Frame()` waits on) tends to drift away from frame start, which is timed by vsync, due to minimal differences, like the display running at 59.95Hz while my timer runs at exactly 60Hz, and probably also due to Window Compositor weirdness.

ALord7:
Do you need to test this? I could join in.

DanielGibson:
I fixed the issues mentioned in my previous comment by mostly getting rid of the async-thread (now it only does sound updates, does **not** update `com_ticNumber` anymore and doesn't get involved in timing in any other way either).

Now it should *really* be smooth, especially on 60Hz displays with VSync enabled.

Updated testbuilds for Windows: https://github.com/dhewm/dhewm3/actions/runs/19418629106?pr=701

Mirroring the Windows build here so you can download it without logging into Github: 
[dhewm3-win32-b96747c.zip](https://github.com/user-attachments/files/23574491/dhewm3-win32-b96747c.zip)

_**Update**, a small clarification: with "Now it should *really* be smooth" I mean that there's no more microstuttering due to timing issues. There are other reasons for Doom3 to shortly stutter like doing a Quicksave or taking a Screenshot or Doom3 updating the objective, which implicitly creates a screenshot. Those stutters persist, but are IMHO rare enough to be tolerable (and also hard to avoid)._

klaussilveira:
Runs pretty smooth on Debian 12, vsync enabled, monitor at 240hz.

```
xrandr
Screen 0: minimum 8 x 8, current 3440 x 1440, maximum 32767 x 32767
DP-0 connected primary 3440x1440+0+0 (normal left inverted right x axis y axis) 1mm x 1mm
   3440x1440    239.99*+ 143.97   119.99    99.98    59.97    49.99
```

This is just as smooth as BFG.

DanielGibson:
As a bonus, the Widescreen GUIs created by @Arl90 for the base game, RoE and Lost Mission: [Wide Guis.zip](https://github.com/user-attachments/files/23585044/Wide.Guis.zip)
They work with all test builds of the last months including the ones linked here.

Just put the .pk4 files from the zip in the corresponding game directories (zWideGuis_D3.pk4 into base/, zWideGuis_RoE.pk4 into d3xp/, zWideGuis_LM.pk4 into d3le/)

markanini:
I tested on a 120hz display, with vsync on. The resulting fps was around 60, not the 62.5 of the regular build. The frame pacing is noticeably improved!

PR #706, opened 2025-12-13
Recalculate gameLocal.msec each frame so 60 frames add up to 1000ms

### LINKED ISSUE #703: Pre-rendered video and audio are out of sync
**Describe the bug**
When you watch an in-game video, audio and video are synchronised when the video begins, but audio plays slightly faster than video, which means that as time passes the audio and the video get more and more out of sync.

**Your System**
- Windows 11
- dhewm3 1.5.4 

**To Reproduce**
Play any video in-game. I've seen it happen only with videos that play on in-game screens. For example, when Kelly calls you in Delta 1. It's possible that it also happens with PDA videos, but I don't know how to check since those do not have people talking in them so it's hard to tell if they are out of sync.

DanielGibson:
based on a patch from @dezo2 for high-fps support, who took the general idea from D3BFG

this works around the problem that for most framerates, with integer frametimes, gameHz * frametime doesn't add up to 1000ms (e.g. 120Hz * 8ms = 960ms), by making some frames a millisecond longer

This problem even exists at 60fps, where the integer frametime is 16 (while it should be 16.666), this causes audio and video in cut scenes (or longer animated talking NPC) and videos to get out of sync. It only worked properly when running without vsync so the game ran at 62.5fps - but now that dhewm3 enforces proper 60fps, the problem always occurs. This commit should fix that.

fix #703

### DIFF
--- a/neo/d3xp/Game_local.cpp
+++ b/neo/d3xp/Game_local.cpp
@@ -2428,11 +2428,11 @@ void idGameLocal::SortActiveEntityList( void ) {
 idGameLocal::RunTimeGroup2
 ================
 */
-void idGameLocal::RunTimeGroup2() {
+void idGameLocal::RunTimeGroup2( int msec_fast ) { // dezo2/DG: added argument for 16 vs 17ms
 	idEntity *ent;
 	int num = 0;
 
-	fast.Increment();
+	fast.Increment( msec_fast );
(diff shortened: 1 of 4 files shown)


DanielGibson:
testbuild for windows: [dhewm3-win32-ff5f9f2.zip](https://github.com/user-attachments/files/24144364/dhewm3-win32-ff5f9f2.zip)

Issue #709, opened 2025-12-21
MSYS2 Clang64 fails to build with curl


thatsprettygood:
**Describe the bug**
Despite reporting that curl has been found, fails with symbol error relating to curl.
When modifying the cmakelists to disable curl, it builds successfully

**Your System**
MSYS2 Clang64
Git source

**Additional information**
I've modified the cmakelists in line 257 to provide information on the curl directory
`message(STATUS "libcurl found and enabled - CURL_LIBRARY = ${CURL_LIBRARY}")`
CMake log:
```
-- Building for: Ninja
-- The C compiler identification is Clang 21.1.7
-- The CXX compiler identification is Clang 21.1.7
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/msys64/clang64/bin/cc.exe - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/msys64/clang64/bin/c++.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- `C:/msys64/clang64/bin/cc.exe -dumpmachine` says: "x86_64-w64-windows-gnu"
--   => CPU architecture extracted from that: "x86_64"
-- Setting -DD3_ARCH="x86_64" -DD3_SIZEOFPTR=8 -DD3_OSTYPE="windows"
-- Found OpenAL: C:/msys64/clang64/lib/libopenal.dll.a
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE
-- Found SDL2: mingw32;C:/msys64/clang64/lib/libSDL2.dll.a;-lpthread
-- Dear ImGui integration enabled
-- libcurl found and enabled - CURL_LIBRARY =
-- Compiler "Clang" detected as some kind of clang
-- Detected Little Endian architecture, setting -DD3_IS_BIG_ENDIAN=0
-- Performing Test cxx_has_fp-contract
-- Performing Test cxx_has_fp-contract - Success
-- Performing Test cxx_has_fvisibility
[...]

PR #710, opened 2025-12-21
Fix compiling on MSYS2

thatsprettygood:
My fix for #709
I'm not familiar with CMake, so sorry if there is a better way to do this


thatsprettygood:
@DanielGibson

DanielGibson:
thanks!

PR #714, opened 2025-12-29
Fix size of threadId field for SDL3 compatibility

carlo-bramini:
I built the engine with SDL3 but I got this error on the console:
```
neo/sys/threads.cpp:61:39: error: static assertion failed: xthreadInfo::threadId has unsuitable type!
   61 |   static_assert( sizeof(SDL_threadID) <= sizeof(xthreadInfo::threadId), "xthreadInfo::threadId has unsuitable type!" );
      |                  ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
neo/sys/threads.cpp:61:39: note: the comparison reduces to ‘(8 <= 4)’
```
This happens because SDL3 handles the thread identifiers with different type than SDL2.

- SDL2 declares SDL_threadID as unsigned long.
- SDL3 declares SDL_threadID as uint64.

So, that's why this trouble happens.
I have seen that `CMakeLists.txt` adds `D3_SDL3` macro to all sources with `add_definitions()` when SDL3 is selected.
This macro allows to check which library is used and adjust the declaration of `threadId` field.
So, I fixed `sys_public.h` with the attached patch and now it works fine with both SDL2 and SDL3.

This is a solution, but it is also possible to set the type to `uint64_t` and ignoring which SDL library is used, if you want to make things even simpler.
However, this is easy to change if someone wants.

I used `aarch64-w64-mingw32` and `x86_64-w64-mingw32` cross compilers for building the engine.

### DIFF
--- a/neo/sys/sys_public.h
+++ b/neo/sys/sys_public.h
@@ -383,7 +383,11 @@ typedef int (*xthread_t)( void * );
 typedef struct {
 	const char		*name;
 	SDL_Thread		*threadHandle;
+#ifdef D3_SDL3
+	uint64_t		threadId;
+#else
 	unsigned long	threadId;
+#endif
 } xthreadInfo;
 
 void				Sys_CreateThread( xthread_t function, void *parms, xthreadInfo &info, const char *name );


DanielGibson:
Hmm apparently I haven't tried building for Win64 with SDL3 yet - thanks!

(good thing I thought of adding that static_assert() =))

DanielGibson:
oh, found another nasty bug, surprising threading code worked at all with SDL3:
In SDL1.2 and SDL2 `SDL_ThreadID()` is a function that returns the current thread's ID.
In SDL3 `SDL_ThreadID` is that typedef to uint64_t. So in C++ `SDL_ThreadID()` is `uint64_t()` which I guess is valid code, but certainly not what we want...
Will push a fix.

see also https://github.com/libsdl-org/SDL/pull/8865

DanielGibson:
see https://github.com/dhewm/dhewm3/commit/f2943c96ced8e139f6e43f2c586fcc4b5ef9f532

Issue #718, opened 2026-01-04
When packaging on linux, it stops on console


Link4Electronics:
Running Arch Linux, when packaging and trying to launch game, it works but stop at console with `unknown command \$@`
Issue relies in `org.dhewm3.Dhewm3.desktop` and `org.dhewm3.Dhewm3.d3xp.desktop`, removing "\\$@" from `Exec` line fixes it.

Regards,
Link.

DanielGibson:
Is this maybe related to #569 ?
pinging @mnhauke 

Generally I don't really support the stuff in `dist/`, it's up to distro package maintainers to install the desktop files etc and make them work - but I'll gladly take patches/pull requests with fixes, of course.

No idea if the Arch Linux package even uses that or writes their own desktop files, you should probably also report the problem to the package maintainer

DanielGibson:
Also, it's probably relevant to know how you launch dhewm3 and what desktop environment you're using

Link4Electronics:
Hi, I just double clicked on desktop icon, using xfce4. `makepkg` finds and grabs `*.desktop` files itself when packaging

DanielGibson:
IIRC these .desktop files were originally written by @Eonfge.
What is the purpose of the `$@` there anyway?

Based on what this means in shell scripts, it would pass arguments given to the .desktop file on to dhewm3 - but **how would the user pass arguments to the .desktop file** in the first place? Certainly not by double-clicking it on the desktop or clicking the entry in the start menu..

Also, I don't see `$@` mentioned in https://specifications.freedesktop.org/desktop-entry/latest/

Eonfge:
Not sure, the Flathub version has a special wrapper-script that is used to check the pk4. Without the pk4 files in the right direction, the Flatpak will not 'start' and it will instead give a notification.

For Arch where that sandboxing and file-checking doesn't exist, the "$@" is not necessary at all.

DanielGibson:
What does $@ do about the pk4s?

Link4Electronics:
I wonder if could be replaced with `%U`? I guess then everyone wins, appimage, flatpak and distro packaging. Theorically should work with flatpak, use of $@ is non-standard and `Exec` line isn't a shell script, it's for direct command execution.

Regards,
Link.

Eonfge:
The flatpak first runs a bash script, to check the file installation before it starts dhewm3.

[...]

PR #719, opened 2026-01-05
Fix for desktop entries

Link4Electronics:
This way it's possible to go to the menu, otherwise when packaging, it would stay at console with uknown command \$@

Fixes https://github.com/dhewm/dhewm3/issues/718


DanielGibson:
thanks!

PR #751, opened 2026-03-24
fixed Crash at air vent in old comm engineering raspberry pi 5

TheMidnightHammer:
just removed this commit `c394fde58a37c2fa4efed39b63d5ff9ac56dad63 `that crashed the game on raspberry pi 5

Fixes #749


DanielGibson:
This PR does not remove or undo that commit but adds a **new** CMakeLists.txt

Furthermore, there might be a better fix than removing these compiler flags, we haven't figured out yet how/why `-ffinite-math-only` breaks `idCollisionModelManagerLocal::Rotation180()`

PR #757, opened 2026-05-08
Pi5 crash fix temporaly

TheMidnightHammer:
This code is not mine it is @DanielGibson who made this, i am just the one to commit them.

This will fix the crash on Raspberry pi 5 temporarily.


DanielGibson:
pushed basically the same changes

